PR #1, opened 2020-02-11
Spike sassdocs integration

nickcolley:
There's a fair bit of complexity in the render view logic which would be nice to simplify but this is hopefully a good proof of concept for us to talk about.

### DIFF
--- a/source/sass_api_reference/index.html.md.erb
+++ b/source/sass_api_reference/index.html.md.erb
@@ -1,0 +1,85 @@
+---
+title: Sass API Reference
+weight: 65
+---
+
+# Sass API Reference
+
+<%# Group the docs by their 'group' %>
+<% groupedDocs = data.sassdoc.group_by{ |doc| doc.group.first } %>
+<% groupedDocs.each do |index, group| %>
+
+## <%= index == 'undefined' ? 'General' : index.gsub('/', ' / ').titlecase %>
+
+    <%# Remove private entries %>
+    <% docs = group.select{ |doc| doc.access == 'public' } %>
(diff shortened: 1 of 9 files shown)


source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,85 @@
...
+        <% end %>
+
+        <% if doc.context.type == 'variable' %>
+```scss
+$<%= doc.type.downcase %>: $<%= doc.context.name %>;
+```
+        <% end %>
+
+        <% if doc.example %>
+#### Example
```
nickcolley: I don't like how the markdown needs to be indented differently to the rest of the file for it to render properly, makes it hard to read...

m-green:
So good to see this in the TDT! Awesome work Nick.
Quick question - is is possible to remove the `<hr>` at the end of each subsection?

nickcolley:
@m-green done

nickcolley:
Source code links should be permalinked to the specific version being deployed

nickcolley:
If we updated tech docs gem to allow `disable_indented_code_blocks: true` we could indent the file and have it read out easily.

nickcolley:
Added a change to handle deprecation warnings:

nickcolley:
https://github.com/alphagov/tech-docs-gem/pull/170

source/sass_api_reference/index.html.md.erb
nickcolley: https://github.com/alphagov/tech-docs-gem/pull/170

PR #15, opened 2020-03-19
Add sass docs

### LINKED ISSUE #12: Publish GOV.UK Frontend sass docs into tech docs format
## What
Publish the Sassdocs from GOV.UK Frontend 

## Why
We know it's not perfect but it's helpful for our users. A few users know about the unofficial https://govuk-frontend-review.herokuapp.com/docs/ and we want to make it official for people to use.

Mark G, Developers, Chris?

## Further detail
A [spike has been done to see if it's possible and it is](https://github.com/alphagov/govuk-frontend-docs/pull/1). We want to clean this work up.

Additional spike: https://github.com/alphagov/govuk-frontend-docs/pull/14

## Depends on
- [ ] A production tech docs instance, currently the prototype exists at: https://github.com/alphagov/govuk-frontend-docs

## Done when

- [x] Add functionality to output sassdocs into tech docs format.
    - [x] GitHub link should refer to current package tag rather than master
- [ ] Add tests to ensure output is correct
- [x] Optional: Add functionality to [allow for blocks in warning_text helper](https://github.com/alphagov/govuk-frontend-docs/pull/1/files#diff-308dd7447f91dcf32c4f62bd2db97c79R1-R8).
- [ ] Raise issues for future sassdocs specific improvements.

## Future GOV.UK Frontend specific improvements
These are improvements to the Sass docs themselves in the main repository.

- [ ] Ask Mark G about improving the sass docs comments so there is no 'general' section for main-wrapper. Might not be an issue if these are removed in 4.0.0...
- [ ] Add examples to relevant variables and functions
- [ ] Check general consistency

nickcolley:
Add the sass docs documentation for GOV.UK Frontend into the public documentation.

Paired on this with @hannalaakso based on the initial spike done to clean it up and make it easier to understand.

Also introduces linting to the project.

Closes https://github.com/alphagov/govuk-frontend-docs/issues/17
Closes #12


nickcolley:
@36degrees we've not added any tests for now as we're not sure what the best way to do this Middleman and Ruby, so wanted to get your opinion on how to proceed.

config.rb
```diff
@@ -1,3 +1,119 @@
...
+        data.sassdoc
+            .select { |doc|
+                # Remove private API entries
+                doc.access == 'public'
+            }
+            .select { |doc|
+                # Remove vendored files, for example SassMQ
+                !doc.file.path.start_with?('vendor')
+            }
+            .map { |doc|
```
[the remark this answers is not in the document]
nickcolley: yeah I like this idea

.gitignore
```diff
@@ -17,4 +17,7 @@
 # Ignore .DS_store file
 .DS_Store
 
+# Node.js / npm
+node_modules
```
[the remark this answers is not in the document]
nickcolley: I figured committing it after each npm install felt right in this case since it'll allow us to check any updates if necessary. Also means we can keep the deployment pipeline simpler.

source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
+---
+title: Sass API Reference
+weight: 65
+---
+
+# Sass API Reference
```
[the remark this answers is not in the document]
nickcolley: Since the markdown generates anchor links and classes I figured using markdown is the most maintainable way for the future, with the compromise of not being able to nest logic.

config.rb
```diff
@@ -1,3 +1,119 @@
 require 'govuk_tech_docs'
 
 GovukTechDocs.configure(self)
+
+# Get the current version of GOV.UK Frontend from the package.
+package_lock_file = File.read('./package-lock.json')
```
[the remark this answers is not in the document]
nickcolley: I didn't put this stuff in a method so it only gets called once, perhaps there is a way to memoise a method in ruby or only call it once?

Can give it a go refactoring but I'm not super hot on modules in Ruby.

config.rb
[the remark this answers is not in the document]
nickcolley: This makes sense

m-green:
Hi @nickcolley - I’ve taken a look through and it looks great! I could only see the folllowing issues:

1. Internal links (for example under ‘Requires’ headings) do not work. It looks like the links are using all the heading levels, rather than just the final heading level.
2. When an item is private (for example #private-govuk-font-face-gds-transport), could we use “(private)” at the end of the heading rather than “[private]” at the start?
3. Several of the the ‘Requires’ list seem to have multiple duplicate links. For example #govuk-typography-responsive has 3 x ‘iff’ links, 2 x ‘govuk-typography-scale’ links, etc. #govuk-visually-hidden-focusable has 16 x ‘iff’s!

nickcolley:
@m-green is your feedback for the currently live developer docs on https://govuk-frontend-review.herokuapp.com/docs/ ? We're not outputting private or 'requires' in this pull request.

m-green:
> @m-green is your feedback for the currently [live developer docs](https://govuk-frontend-review.herokuapp.com/docs/)? We're not outputting private or 'requires' in this pull request:
> 
> https://deploy-preview-15--govuk-frontend-docs-prototype.netlify.com/sass_api_reference/#sass-api-reference

Oh sorry - maybe I'm looking at the wrong version. Hang on, I'll check.

.gitignore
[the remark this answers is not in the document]
nickcolley: Happy to review this decision, might be better to avoid having to commit things, just will need to make sure travis is setup correctly.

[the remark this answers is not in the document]
nickcolley: Going to give this a go...

nickcolley:
@36degrees @hannalaakso I've added tests for all the helper methods, so it's now ready for review again.

sassdocs_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+                # Remove private API entries
+                doc.access == 'public'
+            }
+            .select { |doc|
+                # Remove vendored files, for example SassMQ
+                !doc.file.path.start_with?('vendor')
+            }
+            .group_by {
+                # Group the docs by their 'group', for example 'Settings' or 'Helpers'
+                |doc| doc.group.first
```
36degrees: Minor, but there's inconsistency with the placement of the arguments here – in the two `select` calls the arguments are on the same line as the opening brace.

source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+    <% end %>
+<% end %>
+
+<% if doc.parameter %>
+    <% inline_parameters = "(#{inline_parameters(doc.parameter)})" %>
+<% end %>
+
+<% if doc.context.type == 'mixin' %>
+```scss
+@include <%= doc.context.name %><%= inline_parameters %><%= mixin_trailing_code(doc.context.code) %>
```
36degrees: I found this a bit confusing at first because you've used the same name for the variable as you have for the helper, so at first glance I thought this was calling the helper without any parameters.

I _think_ it might also cause problems if we wanted to use the inline_parameters again in the code, as I assume the variable's existence means there's no longer a way to call it?

sassdocs_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+        #  //..
+        # }
+        acceptsContent = code.include?('@content')
+        if acceptsContent
+            " {\n  //..\n}"
+        else
+            ";"
+        end
+    end
+    def inline_parameters(parameters)
```
36degrees: Do you think it'd make sense to have one helper that's responsible for building the entire example, including any parameters and inline code? It could still call out to private methods, so that we can keep the methods nice and short.

It feels a little strange at the minute as the logic is kind of split, with some of it handled in the helper but some of it being handled in the view.

It'd also mean that we'd have better test coverage, as the logic that's currently composing these in the view can't be tested.

Happy to leave this as is if you'd prefer, and we can refactor it later.

sassdocs_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+        # Generate the possible paramters that could be passed into a function or mixin.
+        # For example with the `govuk-colour` function: `$colour, $legacy`
+        # If the function / mixin has default values output them too, for example `$important: true`
+        parameters.map.with_index { |param, index|
```
36degrees: We don't appear to use the index, so is `with_index` required?

sassdocs_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+                end
+            end
+            output
+        }.join(', ')
+    end
+    def parameters_table(parameters)
+        # Clone the parameters so we don't overwrite the original data when we map over it.
+        parameters.deep_dup.map { |param|
+            # If cell contains no data hide dash from assistive technology
+            no_data = "<span aria-hidden='true'>—</span>"
```
36degrees: Again minor, but can we flip the quotes so we get double quotes on the HTML attributes consistently?

sassdocs_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+                encoded_description = param.description.gsub('|', '&#124;')
+                ## Remove unnecessary line breaks
+                description_without_linebreaks = encoded_description.gsub(/\n/m, '').sub(/\r/m, '')
+                param.description = description_without_linebreaks
+            else
+                param.description = no_data
+            end
+
+            if param.type
+                # Mutliple types can be seperated by pipe characters e.g. "String | Boolean".
```
36degrees:
```suggestion
                # Multiple types can be separated by pipe characters e.g. "String | Boolean".
```

sassdocs_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+        # }
+        acceptsContent = code.include?('@content')
+        if acceptsContent
+            " {\n  //..\n}"
+        else
+            ";"
+        end
+    end
+    def inline_parameters(parameters)
+        # Generate the possible paramters that could be passed into a function or mixin.
```
36degrees:
```suggestion
        # Generate the possible parameters that could be passed into a function or mixin.
```

sassdocs_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+        else
+            'General'
+        end
+    end
+    def github_url(doc)
+        # Construct GitHub link
+        "https://github.com/alphagov/govuk-frontend/tree/v#{govuk_frontend_version}/src/govuk/#{doc.file.path}#L#{doc.context.line.start}-L#{doc.context.line.end}"
+    end
+    def govuk_frontend_version
+        # Since this accesses the file system,
```
36degrees: Neat 👏

source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+
+## <%= group_heading(heading) %>
+
+<% group.each do |doc| %>
+
+### <%= doc_heading(doc) %>
+
+<%= doc.description %>
+
+<% if doc.parameter %>
```
36degrees: Does this table need a heading?

source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+
+<% if doc.context.type == 'function' %>
+```scss
+$result: <%= doc.context.name %><%= inline_parameters %>;
+```
+<% end %>
+
+<% if doc.context.type == 'variable' %>
+```scss
+$<%= doc.type.downcase %>: $<%= doc.context.name %>;
```
36degrees: This doesn't seem quite right, as it outputs e.g. 

```scss
$boolean: $govuk-include-default-font-face;
```

which isn't valid sass, unlike the other usage examples.

source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+
+<% if doc.context.type == 'mixin' %>
+```scss
+@include <%= doc.context.name %><%= inline_parameters %><%= mixin_trailing_code(doc.context.code) %>
+```
+<% end %>
+
+<% if doc.context.type == 'function' %>
+```scss
+$result: <%= doc.context.name %><%= inline_parameters %>;
```
36degrees: I don't know how useful these are – the mixin example seems useful as it stands alone, but both the function and variable examples suffer because they need additional code around them.

Would it be better to try and provide more examples in the doc blocks in Frontend, and rely on them instead?

source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+<% end %>
+
+<% if doc.context.type == 'variable' %>
+```scss
+$<%= doc.type.downcase %>: $<%= doc.context.name %>;
+```
+<% end %>
+
+<% if doc.example %>
+#### Example
```
36degrees: Thoughts on pluralising this, as there may be more than one?

```suggestion
#### Examples
```

source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+<% doc.example.each do |example| %>
+##### <%= example.description %>
+
+```<%= example.type %>
+<%= example.code %>
+```
+<% end %>
+<% end %>
+
+View source code on GitHub: <a href="<%= github_url(doc) %>"><%= doc.file.name %></a>
```
36degrees: Is 'View source code on GitHub' better as the link text for this?

spec/sassdocs_helpers_spec.rb
```diff
@@ -0,0 +1,319 @@
...
+require 'active_support/all'
+
+require_relative '../sassdocs_helpers.rb'
+
+# Middleman helpers convert data to allow dot access, so we need to bring in data in a similar way.
+def dothash(hash)
+  JSON.parse(hash.to_json, object_class: OpenStruct)
+end
+
+RSpec.describe SassdocsHelpers do
```
36degrees: 👏

spec/sassdocs_helpers_spec.rb
```diff
@@ -0,0 +1,319 @@
...
+    allow(File).to receive(:read).and_return('{ "dependencies": { "govuk-frontend": { "version": "1.0.0" } } }')
+    # Include mixin into a test class to allow us to mock File
+    class Test
+      include SassdocsHelpers
+    end
+    @helper = Test.new
+  end
+
+  describe '#format_sassdoc_data' do
+    it "should raise an error if there is no data" do
```
36degrees: I don't know if we follow this elsewhere, but my personal preference is not to use should in the descriptions.

http://www.betterspecs.org/#should

m-green:
> > @m-green is your feedback for the currently [live developer docs](https://govuk-frontend-review.herokuapp.com/docs/)? We're not outputting private or 'requires' in this pull request:
> > https://deploy-preview-15--govuk-frontend-docs-prototype.netlify.com/sass_api_reference/#sass-api-reference
> 
> Oh sorry - maybe I'm looking at the wrong version. Hang on, I'll check.

Sorry, can confirm I was looking at an out-of-date version! All looks great. As discussed, we'll need to look at tweaking the source comments to fix content issues, but let's make that a Q1 thing.

nickcolley:
@36degrees thanks for the great review, lots to improve, I'll get on it.

sassdocs_helpers.rb
nickcolley: Will enable linting in line with tech-docs-gem to sort this.

sassdocs_helpers.rb
nickcolley: Will investigate this.

source/sass_api_reference/index.html.md.erb
nickcolley: Can you explain why it isnt valid Sass please? I agree it's a bit odd given this example but it should be valid Sass.

source/sass_api_reference/index.html.md.erb
nickcolley: I went with `View source code for <FILENAME> on GitHub`

source/sass_api_reference/index.html.md.erb
nickcolley: I may simplify this based on your other comments regardless.

36degrees: Sorry, yes, on a second pass I can see that this is in fact valid 🤦‍♂ My brain could not make sense of it the first time, however, possibly because I didn't expect a variable to be called boolean? Or the assignment felt backwards? Not sure…

nickcolley: I think I was trying to be too clever here, we should improve the Sass docs themselves instead. I'll raise issues on the main repository.

nickcolley:
@36degrees ready for another review

source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,65 @@
...
+# Sass API Reference
+
+<% format_sassdoc_data(data).each do |heading, group| %>
+
+## <%= group_heading(heading) %>
+
+<% group.each do |doc| %>
+
+### <%= doc_heading(doc) %>
+<span class="govuk-secondary-text">(<%= doc.context.type %>)</span>
```
36degrees: As far as I can tell, this class doesn't seem to exist?

source/sass_api_reference/index.html.md.erb
```diff
@@ -0,0 +1,65 @@
...
+#### Parameters
+
+| Name | Description | Type | Default value |
+| ---- | ----------- | ---- | ------------- |
+<% parameters_table(doc.parameter).each do |param| %>
+|  <%= param.name %> | <%= param.description %> | <%= param.type %> | <%= param.default_value %> |
+<% end %>
+<% end %>
+
+<% if doc.deprecated || doc.context.type == 'mixin' || doc.example %>
```
36degrees: This is a little strange as it can mean you have a h4 usage heading followed immediately by a h4 examples heading, if you have examples but it's not a mixin or deprecated.

https://deploy-preview-15--govuk-frontend-docs-prototype.netlify.com/sass_api_reference/#govuk-spacing

source/sass_api_reference/index.html.md.erb
nickcolley: I think I accidentally committed this, oops thanks for catching.

source/sass_api_reference/index.html.md.erb
nickcolley: agreed will fix

nickcolley:
@36degrees I need to sort the commit history but let me know if you're happy otherwise.

PR #20, opened 2020-04-07
Improve install documentation

m-green:
This updates the Frontend npm install documentation to improve the content's readability and accuracy, based on the findings from our research. I've:

- split the install guide into 1) install, 2) get started and 3) importing
- kept things as simple and linear as possible in 'Get started', but linked to other options in the 'Importing' page
- used a 'try-then-test' approach for 'Get started', to better match the user behaviour we saw
tweaked language in headings and copy to help convey that steps are not optional

The content itself doesn't need reviewing - it's already been 2i'd by technical writers, and reviewed by the team in the Frontend repo.

After this is merged:
- [ ] Update new migration documentation to fix links to this content


source/get-started/index.html.md.erb
```diff
@@ -0,0 +1,95 @@
+---
+title: Get started
+weight: 11
```
36degrees, line 3: It seems a little strange to have this between the two alternate 'Install' sections – would it make sense to have this after them, like:

> GOV.UK Frontend
> Install with Node.js package manager (npm)
> Install from dist
> Get started
> Update using Node.js package manager (npm)
> Support Internet Explorer 8
> Use with old frameworks or colours
> Import CSS, assets and JavaScript
> Sass API Reference

source/get-started/index.html.md.erb
```diff
@@ -0,0 +1,95 @@
...
+
+1. Install GOV.UK Frontend.
+2. Add the HTML for a component to your application.
+3. Get the CSS working.
+4. Get the font and images working.
+5. Get the JavaScript working.
+
+## 1. Install
+
+[Install GOV.UK Frontend using npm](/installing_with_npm).
```
36degrees, line 20: It'd be great to rename the folders so that they're all kebab case, at the minute we have a mix of kebab case and snake case. Possibly makes sense to do this in a follow up PR? I don't mind doing it, if you prefer.

36degrees: Also, again we can avoid a 301 redirect by adding the trailing slash:

```suggestion
[Install GOV.UK Frontend using npm](/installing_with_npm/).
```

source/installing_with_npm/index.html.md.erb
```diff
@@ -26,289 +24,16 @@ To use GOV.UK Frontend with NPM you must:
...
-    <button class="govuk-button">This is a button component</button>
-    <script src="assets/application.js"></script>
-  </body>
-</html>
-```
+When the installation finishes, the `govuk-frontend` package will be in your `node_modules` folder.
 
-If your service supports Internet Explorer 8, you will need to [generate and
-include a separate stylesheet](/supporting_ie8#writing-styles-that-target-ie8) as well.
+You should now get started by [getting the CSS, assets and JavaScript working](/get-started) with one GOV.UK Frontend component.
```
36degrees, line 39: _Super_ minor, but this'll make a request to `/get-started` which will immediately redirect to a `/get-started/` with a trailing slash. From a performance point of view, it's cleaner to add the trailing slash as it saves a request!

```suggestion
You should now get started by [getting the CSS, assets and JavaScript working](/get-started/) with one GOV.UK Frontend component.
```

m-green: Oh wow, I did not know this! Makes me wonder if we've been serving 301s on Pay docs for years. Thanks for letting me know - will fix across the docs.

source/get-started/index.html.md.erb
m-green: Ah yes - sorry that's my mistake. You're right they should all be kebab case. I'll fix that up.

Will fix the 301s too!

source/get-started/index.html.md.erb
m-green: Good plan - will fix in the next commit.

source/installing_with_npm/index.html.md.erb
36degrees: Yep, it's a minor thing that's not worth worrying about too much, but good to build habits. Note also that e.g. GOV.UK is the exact opposite, and redirects to _remove_ the trailing slash!

I only noticed because the Chrome extension I used to check links highlights links that redirect, which makes it easier to spot.

source/get-started/index.html.md.erb
36degrees: We're now consistent, but seem to be using snake_case not kebab-case.

I guess this is as much personal preference as anything, and I'm happy to defer to you on this one, but my personal preference would be for kebab-case, because:

- it's consistent with the way that we slug anchors (e.g. `/importing-css-assets-and-javascript/#font-and-image-assets` vs `/importing_css_assets_and_javascript/#font-and-image-assets`)
- it's consistent with the Design System (https://design-system.service.gov.uk/patterns/contact-a-department-or-service-team/)
- I think it's generally best practise? (Google recommends it anyway – https://support.google.com/webmasters/answer/76329!)

m-green: Thanks Olli - not my day today. Everything should be kebab-case now. I even found one link where I'd used snake_case wrongly in a link to the Design System website!

Issue #26, opened 2020-04-14
Set up 404 and generic error pages


hannalaakso:
## What
Users should see a 404 page on trying to access a non-existent page or a generic error page with other errors

## Why
To avoid confusion

## Who needs to know about this
Devs

## Done when
- [x] A 404 page is shown when accessing `/potato`

PR #33, opened 2020-04-15
Add 404 and generic error pages

hannalaakso:
This adds a 404 and a generic error page. The content of the pages follows what we use in the Design System website.

We include `index: false` in frontmatter to remove the pages from the site search.

I've tested the 404 page by pushing to the sandbox to see it gets picked up there.

Fixes https://github.com/alphagov/govuk-frontend-docs/issues/26

### DIFF
--- a/source/errors/404.html.md.erb
+++ b/source/errors/404.html.md.erb
@@ -1,0 +1,15 @@
+---
+title: Page not found
+hide_from_sitemap: true
+hide_in_navigation: true
+index: false
+layout: core
+---
+
+# Page not found
+
+If you typed the web address, check it is correct.
+
+If you pasted the web address, check you copied the entire address.
+
+[Contact the Design System team](https://design-system.service.gov.uk/get-in-touch/) if you believe you are seeing this message in error.
--- a/source/errors/generic.html.md.erb
+++ b/source/errors/generic.html.md.erb
@@ -1,0 +1,13 @@
+---
+title: Sorry, there is a problem with the service
+hide_from_sitemap: true
+hide_in_navigation: true
+index: false
+layout: core
+---
+
+# Sorry, there is a problem with the service.
+
+Try again later.
+
+[Contact the Design System team](https://design-system.service.gov.uk/get-in-touch/) if you believe you are seeing this message in error.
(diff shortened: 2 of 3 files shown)


source/errors/404.html.md.erb
```diff
@@ -0,0 +1,14 @@
+---
+title: 404 Page not found
```
36degrees, line 2:
```suggestion
title: Page not found
```

Should just be 'Page not found', as per the page not found pattern.

source/errors/404.html.md.erb
```diff
@@ -0,0 +1,14 @@
...
+hide_from_sitemap: true
+hide_in_navigation: true
+index: false
```
36degrees, line 5: Would it make sense to use `layout: core` here so that the error pages don't get the navigation?

source/errors/generic.html.md.erb
```diff
@@ -0,0 +1,12 @@
+---
+title: Sorry, there is a problem with the service
+hide_from_sitemap: true
+hide_in_navigation: true
+index: false
```
36degrees, line 5: Again, would it make sense to use layout: core here so that the error pages don't get the navigation?

source/errors/404.html.md.erb
hannalaakso: This matches the Design System [404 page title](https://github.com/alphagov/govuk-design-system/blob/311350fa50bce0a087760ba49a0e38885f0f5550/src/errors/404.njk#L2). Should we change that as well? It does sound a bit odd.

source/errors/404.html.md.erb
hannalaakso: Good shout 👍

source/errors/generic.html.md.erb
hannalaakso: 👍

source/errors/404.html.md.erb
36degrees: Yep, I reckon so 👍 The pattern in the Design System says 'Do not use […] technical jargon like 404 or bad request'.

hannalaakso: https://github.com/alphagov/govuk-design-system/pull/1232

Issue #27, opened 2020-04-14
Set up LogIt logging


hannalaakso:
## What
We should send our logs to Logit so that we can use Kabana to generate reports.

## Why
We already do this for the Design System website so we should be consistent with our logging.

## Who needs to know about this
Devs, Kelly

## Done when
- [x] Logit receives logs
- [x] Set up some visualisations in Kabana (similar to the Design System website)

PR #38, opened 2020-04-17
Bind logit-ssl-drain service to app

36degrees:
Send logs to the existing logit-ssl-drain service we use for the Design System, by binding the service to the app.

This means that the logs will end up in the same Logit instance as the Design System logs, but we can easily filter the logs and the dashboards by app name.

Part of #27

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -7,6 +7,11 @@ applications:
   # https://docs.cloud.service.gov.uk/deploying_apps.html#how-to-use-custom-buildpacks
   buildpack: https://github.com/cloudfoundry/nginx-buildpack.git#v1.1.7
 
+  # Send application logs to logit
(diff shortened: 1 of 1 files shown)


tlwr:
The `logit-ssl-drain` service exists (https://admin.cloud.service.gov.uk/organisations/a27b3e36-e041-464d-85a9-c9bef5201c97/spaces/a13048ea-5ac1-4ff1-8aa3-d6a2e2e8be22/services) and the syntax is correct

:shipit:

36degrees:
For future reference, because the `logit-ssl-drain` service exists in a different space, we had to set up a _second_ `logit-ssl-drain` service in the frontend-docs space (using the same logit endpoint) in order to be able to bind it.

PR #43, opened 2020-04-28
Add missing script tags to `js-enabled` code block

36degrees:


source/get-started/index.html.md.erb
```diff
@@ -67,8 +67,8 @@ In your live application, we recommend [using an automated task or your build pi
 
 1. Add the following JavaScript to the top of the `<body>` section of your page template:
```
m-green, line 68:
```suggestion
1. Add the following to the top of the `<body>` section of your page template:
```

@36degrees Looks good but suggest we remove 'JavaScript' here now, because technically this is now adding a line of html (with some JavaScript in it).

36degrees: I considered that – think it's _probably_ valid to refer to it like this, because the effect of your action is still to 'add the JavaScript to the top of the body section', but I know what you mean.

PR #46, opened 2020-05-04
Improve 'install from dist' documentation

m-green:
This improves the 'install from `dist` documentation, because it was one of the pieces of documentation we migrated to the Tech Docs Template without improvement.

It:
- improves the style and made it more readable
- improves the HTML code example
- changes the title so it's easier to understand (and adds a redirect from the old url)
- fixes a link in the npm install documentation


source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+If you install this way, you will not be able to:
+
+- change settings, for example you will not be able to override colours or set your own font
+- use the Nunjucks macros from the [Design System website](https://design-system.service.gov.uk/)
+- import an individual component’s CSS or JavaScript
+- use GOV.UK Frontend's colours or mixins in your custom code
+
+## Copy the files
+
+1. Download the [`dist` folder and its subfolders](https://github.com/alphagov/govuk-frontend/tree/master/dist) from the GOV.UK Frontend GitHub repo.
```
36degrees, line 24: I don't know why I didn't spot this before, but we attach a ZIP to the latest release with the contents of the dist folder in it. It'd make more sense to link to there?

https://github.com/alphagov/govuk-frontend/releases

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+2. Copy the 2 `.css` files to `<YOUR-APP>/stylesheets`.
```
36degrees, line 25: Should we consider changing the structure of `dist` so that you everything is already neatly split into `stylesheets`, `javascript` and `assets` folders? (or create an issue to do so in the future)

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+3. Copy the `.js` file to `<YOUR-APP>/javascript`.
+4. Copy the `assets` folder to the root of your project’s public folder, so that for example `<YOUR-SITE-URL>/assets/govuk-logotype-crown.png` will work in your users’ browsers.
```
36degrees, line 27: This is true for all three folders.

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+
+## Check it works
+
+1. Create a page in your project using the following HTML:
+
+    ```html
+    <!DOCTYPE html>
+    <html lang="en" class="govuk-template ">
+      <head>
+        <title>Example if you’ve installed GOV.UK Frontend by copying files</title>
```
36degrees, line 37: Should this use something that fits the Page Title we want people to use?

```suggestion
        <title>Example - GOV.UK</title>
```

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
+---
+title: Install by copying files
+weight: 10
+---
+
+# Install by copying files
+
+You can install GOV.UK Frontend by copying our CSS and JavaScript files into your project.
+
+<%= warning_text('We recommend <a href="/installing-with-npm/">installing with Node.js package manager (npm)</a> instead.
```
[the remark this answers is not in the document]
36degrees: If we wanted to keep the warning concise, perhaps we could add something to the previous line? For example,

'You can install GOV.UK Frontend by copying our CSS and JavaScript files into your project. This is an easy way to try GOV.UK Frontend in your application without having to make many changes.'

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+') %>
+
+## What you cannot do
+
+If you install this way, you will not be able to:
```
36degrees, line 15:
```suggestion
If you install using pre-compiled Sass and JavaScript, you will not be able to:
```

?

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+
+    In your live application, you should use the [page template](https://design-system.service.gov.uk/styles/page-template/) instead.
+
+2. Replace `<VERSION-NUMBER>` so the 3 filenames match the files you copied from GOV.UK Frontend's GitHub repo.
+
+3. Go to the [example accordion component](https://design-system.service.gov.uk/components/accordion/#accordion-example) on the Design System website, then copy the HTML.
+
+4. Replace `<!-- component HTML -->` with the accordion HTML you copied.
+
+5. Run your application and check it works the same way as the Design System accordion example, by selecting the buttons and checking the accordion shows and hides sections.
```
36degrees, line 67: Because the HTML doesn't contain any of the other 'boilerplate' layout, you end up with something that looks like this:

image

Do we think that's OK? Do we need to do anything to set expectations about what they should see, or make it clearer what the next steps would be to get something that looks more like a 'complete' page?

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+        </script>
+      </body>
+    </html>
+    ```
+
+    In your live application, you should use the [page template](https://design-system.service.gov.uk/styles/page-template/) instead.
+
+2. Replace `<VERSION-NUMBER>` so the 3 filenames match the files you copied from GOV.UK Frontend's GitHub repo.
+
+3. Go to the [example accordion component](https://design-system.service.gov.uk/components/accordion/#accordion-example) on the Design System website, then copy the HTML.
```
36degrees, line 63: Do we need to be clear about which HTML they should be copying (not that it really matters?)

```suggestion
3. Go to the [example accordion component](https://design-system.service.gov.uk/components/accordion/#accordion-example) on the Design System website, then copy the HTML from the first example.
```

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
+---
+title: Install by copying files
+weight: 10
+---
+
+# Install by copying files
+
+You can install GOV.UK Frontend by copying our CSS and JavaScript files into your project.
```
hannalaakso, line 8: Also the `assets` folder.

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+
+<%= warning_text('We recommend <a href="/installing-with-npm/">installing with Node.js package manager (npm)</a> instead.
+') %>
+
+## What you cannot do
+
+If you install this way, you will not be able to:
+
+- change settings, for example you will not be able to override colours or set your own font
```
hannalaakso, line 17: Non-blocking because I'm not sure I've got a point... but is it clear that we're talking about doing things in Sass when we talk about 'not being able to override colours'? You can of course do what you like in your CSS file like overriding the colour of a Design System class (even if it's probably not very good practise to do that), maybe if you just wish to test something quickly.

source/install-by-copying-files/index.html.md.erb
hannalaakso: Yeah that makes sense to me, maybe a separate issue?

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+        <!--[if !IE 8]><!-->
+          <link rel="stylesheet" href="/stylesheets/govuk-frontend-<VERSION-NUMBER>.min.css">
+        <!--<![endif]-->
+        <!--[if IE 8]>
+          <link rel="stylesheet" href="/stylesheets/govuk-frontend-ie8-<VERSION-NUMBER>.min.css">
+        <![endif]-->
+      </head>
+      <body class="govuk-template__body ">
+        <script>
+          document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
```
hannalaakso, line 48: Should we make it clearer that this line should stay right after opening `<body>` tag? We might need to use JavaScript with something that's not a component. It might also be helpful for user to know what to do with this line once they copy more Design System HTML into their file.

source/install-by-copying-files/index.html.md.erb
hannalaakso: Or
> copy the HTML for the component.

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
...
+        <!-- component HTML -->
+        <script src="/javascript/govuk-frontend-<VERSION-NUMBER>.min.js"></script>
+        <script>
+          window.GOVUKFrontend.initAll()
+        </script>
+      </body>
+    </html>
+    ```
+
+    In your live application, you should use the [page template](https://design-system.service.gov.uk/styles/page-template/) instead.
```
hannalaakso, line 59: Or even if they just want to get a complete page with header and footer for testing.

source/install-by-copying-files/index.html.md.erb
```diff
@@ -0,0 +1,73 @@
+---
+title: Install by copying files
```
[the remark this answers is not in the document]
m-green: Makes sense - I'll use 'Install using pre-compiled files' for brevity if that's ok.

source/install-by-copying-files/index.html.md.erb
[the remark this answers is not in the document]
m-green: Good idea. If we added that, would it be more accurate for the warning to say "In your live application, you should install with npm"?

source/install-by-copying-files/index.html.md.erb
m-green: Good plan - I've tweaked.

source/install-by-copying-files/index.html.md.erb
m-green: Ah - sorry I might have misunderstood an earlier comment. So we could maybe just say "Copy the `assets` folder to '<YOUR-APP>/", and lose the second bit?

source/install-by-copying-files/index.html.md.erb
m-green: 👍

source/install-by-copying-files/index.html.md.erb
m-green: Good suggestion - but actually, looking at this fresh, I think we can just start with "You will not be able to:"

source/install-by-copying-files/index.html.md.erb
m-green: Ah good point - I think if we use your suggestion of making it clearer what the next steps are, it should probably be ok.

I'll add the same line that we have at the end of 'Getting started': "You can now get the full code for page layouts and other components from the [Design System website](https://design-system.service.gov.uk/)."

source/install-by-copying-files/index.html.md.erb
m-green: Good idea! I've used 'first example' to be super-clear.

source/install-by-copying-files/index.html.md.erb
m-green: Good point! Fixed.

source/install-by-copying-files/index.html.md.erb
m-green: That's a good call - I'll tweak to "change Sass settings"

source/install-by-copying-files/index.html.md.erb
m-green: Ollie raised the same question in our comments doc - I've covered it with the line at the bottom that directs users having trouble to the 'Import CSS, assets and JavaScript' guide, which mentions the positioning.

I'm sorta 70-80% confident about it though. So we should keep an eye on support to see if users have issues later.

source/install-by-copying-files/index.html.md.erb
m-green: Ah yes - I played around with actually replacing this first step with 'copy the page template HTML'. But I think it makes the following steps a bit more complicated, so I think it might be best not to suggest users use the template at this point?

m-green:
@36degrees @hannalaakso Thank you for the great comments! I've committed with tweaks so far - Ollie, there are a couple of remaining questions for you in the comments.

source/install-by-copying-files/index.html.md.erb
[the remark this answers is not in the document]
36degrees: Yep, I think so :+1:

source/install-by-copying-files/index.html.md.erb
36degrees: The documentation for `assets` is the clearest at the minute I think, as there's no guarantee that `<YOUR-APP>/stylesheets` would be served at `/stylesheets` in the browser (in fact it's pretty unlikely).

source/install-by-copying-files/index.html.md.erb
36degrees: Still pending, I think.

source/install-by-copying-files/index.html.md.erb
hannalaakso: I've got some doubts about this as well, we say in step 5
> Run your application and check it works the same way as the Design System accordion example

That to me sounds like it should "look" the same as well. I would probably think I've done something wrong if I saw a full page version of the accordion, it looks broken to me 🤔 

I wonder if we can try to manage expectations by telling them more explicitly what they should expect to see.

m-green: Could we link them to https://design-system.service.gov.uk/components/accordion/default/index.html to compare to that instead maybe?

source/install-by-copying-files/index.html.md.erb
hannalaakso: Thanks @m-green. I don't think we should suggest they use the template, but I wonder if instead of 'live application' this could say something about getting a full page, even if it's for testing.
> To get the full page template with header and footer components, you should use ...

source/install-by-copying-files/index.html.md.erb
hannalaakso: @m-green Yes that's a good idea 👍

source/install-by-copying-files/index.html.md.erb
m-green: Oh sorry I misunderstood your comment - will fix.

source/install-by-copying-files/index.html.md.erb
m-green: We could, but then it might sound like an alternate option in this step - and then they'll hit trouble from step 2 onwards, because they'll need to do more with the full page template to get it working (that we don't take them through here).
I might be misunderstanding what you mean though, so let me know!

hannalaakso: Thanks @m-green. I see your point about it possibly looking like an alternative step. Let's go with your suggestion and see if it comes up on support.

m-green:
@hannalaakso @36degrees Updated with more tweaks! Let me know if you spot anything that still needs fixing.

hannalaakso:
@m-green There are references to "installing from `dist`" or "distributed versions" in Get started, Use with our old frameworks or colours,  Support Internet Explorer 8, Migrate from our old frameworks so the wording there will need to be updated to reflect the one here.

m-green:
> @m-green There are references to "installing from `dist`" or "distributed versions" in Get started, Use with our old frameworks or colours, Support Internet Explorer 8, Migrate from our old frameworks so the wording there will need to be updated to reflect the one here.

Good spot @hannalaakso! Those should all be fixed now.

source/supporting-ie8/index.html.md.erb
```diff
@@ -17,9 +17,7 @@ then you will need to do some additional work to support Internet Explorer 8
 Once you have completed these steps, you will be able to [write CSS that targets
 IE8 in your own application styles](#writing-styles-that-target-ie8).
 
-If you are using the distributed versions of GOV.UK Frontend that already
-include an IE8-specific stylesheet then you can include that in your project,
-allowing you to skip steps 2 and 3.
+If you [installed using precompiled files](/install-using-precompiled-files), you can include the IE8-specific stylesheet in your project, allowing you to skip steps 2 and 3.
```
hannalaakso, line 20: This is the only page that mentions [installing using precompiled files](/install-using-precompiled-files) AND links to it. Could we link to the guidance from the other pages too to help users understand what we mean? Seeing "installing using precompiled files" without a link might sound a bit confusing and make users question whether it's something they've done or not.

m-green: I think the main places we mention precompiled files without a link is where we warn users from using them (for example in migrating or using compatibility mode). So it's probably best to avoid a link there, to avoid users potentially heading off and doing a thing they shouldn't.

I may have missed some mentions though - let me know if so!

hannalaakso: Okay happy to go with your hunch here 👍

m-green:
@36degrees @hannalaakso There are some minor changes from 2i in https://github.com/alphagov/govuk-frontend-docs/pull/46/commits/a062150c79c2bc3a2372ed14ed56a511dc0e4a59 - could you take a look to check nothing's been made factually incorrect?

source/install-using-precompiled-files/index.html.md.erb
```diff
@@ -10,26 +10,26 @@ You can install GOV.UK Frontend by copying our CSS, JavaScript and asset files i
...
 - use GOV.UK Frontend's colours or mixins in your custom code
 
 ## Copy the files
 
 1. Download the `release-<VERSION-NUMBER>.zip` file at the bottom of the [latest GOV.UK Frontend release note](https://github.com/alphagov/govuk-frontend/releases/latest).
 2. Unzip the zip file.
-3. Copy the `assets` folder to the root of your project’s public folder, so that for example `<YOUR-SITE-URL>/assets/govuk-logotype-crown.png` will work in your users’ browsers.
-4. Copy the 2 `.css` files to a stylesheets folder in the root of your project’s public folder, so that for example `<YOUR-SITE-URL>/stylesheets/govuk-frontend-<VERSION-NUMBER>.min.css` will work in your users’ browsers.
-5. Copy the `.js` file to a JavaScript folder in the root of your project’s public folder, so that for example `<YOUR-SITE-URL>/javascript/govuk-frontend-<VERSION-NUMBER>.min.js` will work in your users’ browsers.
+3. Copy the `assets` folder to the root of your project’s public folder, so that for example `<YOUR-SITE-URL>/assets/govuk-logotype-crown.png` shows the `govuk-logotype-crown.png` image in your users’ browsers.
```
36degrees, line 26: Factually correct, so feel free to ignore this, but could it just be ‘a browser’?

I think talking about the user’s browser is confusing… it’s not like you’re going to ask a user to check, and the user would never actually go and look at the CSS or JS.

36degrees:
Need to look at updating https://github.com/alphagov/govuk-frontend#2-install-by-using-compiled-files to reflect these changes.

m-green:
@36degrees Got an interesting issue in `tech-docs.yml' with the redirect on line 43. I can't use a relative link to the destination because it doesn't seem to construct the url correctly. But an absolute links fails the link check, because the destination page isn't live yet... Any ideas?

PR #53, opened 2020-05-19
Add a link checker and fix broken links

### LINKED ISSUE #8: Add a broken link checker
## What

We should test our generated markup for broken links. 

## Why

To ensure that any broken links are flagged up as part of the build.

## Extra detail

We should probably solve for this the tech docs template rather than just this repo

## Done when…

- [ ] We've added a broken link checker which fails the build if broken links are found

36degrees:
You can see an example failing build [here](https://travis-ci.com/github/alphagov/govuk-frontend-docs/builds/166856383#L749-L765):

```
Guessing --root from input files: file:///home/travis/build/alphagov/govuk-frontend-docs/deploy/public/
..................×..........................................................................................................................................................................................................................................................................................................................×........................................................................................................................................................................................................................................................................  
  ✖ FAIL load deploy/public/installing_with_npm
  | operator: load
  | expected: 200 deploy/public/installing_with_npm
  |   actual: ENOENT: no such file or directory, open '/home/travis/build/alphagov/govuk-frontend-docs/deploy/public/installing_with_npm'
  |       at: deploy/public/compatibility-mode/index.html:304:108 <a href="/installing_with_npm">...</a>
  
  ✖ FAIL fragment-check deploy/public/importing-css-assets-and-javascript/index.html --> /compatibility-mode/#use-with-old-frameworks-or-colours
  | operator: fragment-check
  | expected: id="use-with-old-frameworks-or-colours"
  |       at: deploy/public/importing-css-assets-and-javascript/index.html:315:208 <a href="/compatibility-mode/#use-with-old-frameworks-or-colours">...</a>
  
  598 tests
  596 passed
    2 failed
```

Closes #8

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,11 +5,14 @@
   "scripts": {
     "postinstall": "npm run build:sassdoc",
     "build:sassdoc": "sassdoc --parse node_modules/govuk-frontend/govuk/ > data/sassdoc.json",
-    "lint": "standard"
+    "lint": "standard",
+    "check-links": "hyperlink -ri deploy/public/index.html | tap-spot"
   },
   "devDependencies": {
     "govuk-frontend": "^3.6.0",
+    "hyperlink": "^4.5.0",
     "sassdoc": "^2.7.1",
-    "standard": "^14.3.3"
+    "standard": "^14.3.3",
+    "tap-spot": "^1.1.1"
   }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ cache: bundler
 script:
   - bundle exec middleman build
   - bundle exec rake
+  - npm run check-links
 
 env:
   global:
--- a/README.md
+++ b/README.md
@@ -70,7 +70,9 @@ Run `bundle exec middleman build --verbose` to get detailed error messages to he
 
 ## Testing
 
-Run `bundle exec rake`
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -7,11 +7,14 @@
     "build:sassdoc": "sassdoc --parse node_modules/govuk-frontend/govuk/ > data/sassdoc.json",
     "scripts": {
       "lint": "standard"
-    }
+    },
+    "test": "hyperlink -ri deploy/public/index.html | tap-spot"
```
hannalaakso, line 11: Just a suggestion but could this be
```
"test": "bundle exec middleman build && hyperlink -ri deploy/public/index.html | tap-spot"
```
so that you don't need to run the build separately to see any errors? Travis could also then just run `npm test`.

hannalaakso, line 11: Would it be possible to document what the `-ri` option does? I can't find it in the library docs.

36degrees:
> Should this be checking for external broken links as well? Noticed that something like `https://www.gov.uk/sevice-manual/` doesn't get flagged.

I've kept it to internal only because, for example, the Sass API reference links out to the source code on GitHub for every item in the API, which means we hit GitHub a lot and the tests take a long time to run.

It could also cause unrelated PRs to fail if e.g. an external site changed its URL structure.

I guess we should revisit this if we find broken links to external sites commonly occurring in the future?

PR #55, opened 2020-05-20
Fix linting and run tests on Travis

hannalaakso:
This PR:
- Moves linting to scripts so it can be run (Rakefile wasn't able to find the linter with `npm run lint` and the error was suppressed)
- Removes suppressing linting errors when tests are run from Rakefile
- Fixes violations picked up by Rubocop
- Runs tests on Travis
- Excludes the `vendor` folder from Rubocop because on Travis it contains the other gems which can have configs that conflict with `rubocop-govuk` which we use for our code

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,12 @@ inherit_gem:
   rubocop-govuk:
     - config/default.yml
 
+# Exclude `/vendor` because on Travis other gems get installed there
+# Not a problem locally as gems get installed globally outside project 
+AllCops:
+  Exclude:
+    - './vendor/**/*'
+
 Style/HashEachMethods:
   Enabled: false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,9 @@ language: ruby
 # shipped with Ruby 2.5.
 dist: bionic
 cache: bundler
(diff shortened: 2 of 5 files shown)


.rubocop.yml
```diff
@@ -2,6 +2,10 @@ inherit_gem:
   rubocop-govuk:
     - config/default.yml
 
+AllCops:
```
36degrees, line 5: Is it worth adding a comment that this is needed for Travis, do you think?

.travis.yml
```diff
@@ -4,7 +4,9 @@ language: ruby
 # shipped with Ruby 2.5.
 dist: bionic
 cache: bundler
-script: bundle exec middleman build
+script:
+  - bundle exec middleman build
+  - npm test
```
36degrees, line 9: I think I'd prefer to avoid the indirection and just run `bundle exec rake` from here, and not add it to `package.json` – either way it's a bit weird as we're mixing ruby and node tooling, but wrapping the ruby tooling in node tooling seems to add to the confusion rather than simplifying things. What do you think?

hannalaakso: Yes that makes sense to me too 👍

PR #56, opened 2020-05-21
Update guidance on importing parts of Frontend

### LINKED ISSUE #47: Update documentation to reference new ways of importing components
## What

- Update the guidance on [importing CSS, assets and JavaScript](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#import-an-individual-component-39-s-css) to reflect the changes in alphagov/govuk-frontend#1797.

- Check if there are other parts of the documentation that also need to be updated.

## Further detail

With the changes in alphagov/govuk-frontend#1797, the preferred way to import a subset of the components will be to do something like:

```scss
@import "govuk/base";

@import "govuk/components/button/index"; 
@import "govuk/components/details/index"; 
```

If using a Sass compiler [that supports index files](https://sass-lang.com/documentation/at-rules/import#index-files), this could just be:

```scss
@import "govuk/base";

@import "govuk/components/button"; 
@import "govuk/components/details"; 
``` 

You can still import a single component without having to separately import its dependencies:

```scss
@import "govuk/components/button/button";
````

However, this is only really meant to be used when you need to import a single component in isolation, e.g. if you're creating a React component.

## Why

Importing the components using the index file avoids re-importing the 'base' (settings, tools and helpers layers) multiple times, improving compilation times.

@m-green, @36degrees, @hannalaakso

### PR DESCRIPTION
**Pull request by 36degrees**
With the changes in GOV.UK Frontend v3.7.0, the preferred way to import a subset of the components has changed, with the introduction of a new index file for each component.

Importing the components using the index file avoids re-importing the 'base' (settings, tools and helpers layers) multiple times, improving compilation times.

Differentiate between importing _select bits_ of GOV.UK Frontend vs importing a single specific component.

Include `/index` by default when importing, as older Sass compilers don't support importing index files, but add a note that explains that you can remove it if you are using a newer version.

Closes #47

### DIFF
--- a/source/importing-css-assets-and-javascript/index.html.md.erb
+++ b/source/importing-css-assets-and-javascript/index.html.md.erb
@@ -15,34 +15,36 @@ To import all the Sass rules from GOV.UK Frontend, add the following to your Sas
 @import "node_modules/govuk-frontend/govuk/all";
 ```
 
-## Import only the parts of GOV.UK Frontend that you need
+### Import specific parts of the CSS
 
-In your Sass file, import `base` followed by the parts of GOV.UK Frontend that you need. For example:
+If you want to improve how quickly Sass compiles to CSS, you can import only the Sass rules you need.
+ 
+1. Import `node_modules/govuk-frontend/govuk/base` in your Sass file.
+2. Import the parts of the CSS you need.
+
+For example, add the following to your Sass file to import the CSS you need for a basic GOV.UK page.
 
 ```scss
 @import "node_modules/govuk-frontend/govuk/base";
 
-// Typography, link styles, etc.
-@import "node_modules/govuk-frontend/core/all";
-
-// Grid and layout
-@import "node_modules/govuk-frontend/objects/all";
-
-// Specific components
-@import "node_modules/govuk-frontend/components/footer/index";
-@import "node_modules/govuk-frontend/components/header/index";
-@import "node_modules/govuk-frontend/components/skip-link/index";
-
-// Clearfix and visually hidden utilities
-@import "node_modules/govuk-frontend/utilities/all";
-
-// Overrides for spacing, typography, etc.
-@import "node_modules/govuk-frontend/overrides/all";
+@import "node_modules/govuk-frontend/govuk/core/all";
+@import "node_modules/govuk-frontend/govuk/objects/all";
+@import "node_modules/govuk-frontend/govuk/components/footer/index";
+@import "node_modules/govuk-frontend/govuk/components/header/index";
+@import "node_modules/govuk-frontend/govuk/components/skip-link/index";
+@import "node_modules/govuk-frontend/govuk/utilities/all";
+@import 
+"node_modules/govuk-frontend/govuk/overrides/all";
 ```
+You can remove lines that import parts of the CSS you do not need. Read more about [the different parts of GOV.UK Frontend’s CSS](https://github.com/alphagov/govuk-frontend/tree/master/src/govuk).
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by m-green** [comment]
@36degrees Updated following our discussion! Let me know if it looks ok to go to 2i.

File: source/importing-css-assets-and-javascript/index.html.md.erb [hunk 1]
```diff
@@ -15,9 +15,36 @@ To import all the Sass rules from GOV.UK Frontend, add the following to your Sas
 @import "node_modules/govuk-frontend/govuk/all";
 ```
 
-### Import an individual component's CSS
+### Import specific parts of the CSS
 
-To import the [button](https://design-system.service.gov.uk/components/button/) component for example, add the following to your Sass file:
+If you want to improve how quickly Sass compiles to CSS, you can import only the Sass rules you need.
```
36degrees, line 20: This will also (and arguably a lot more importantly) improve performance for end users, as the compiled CSS will only include the bits required for the service, so it'll be smaller, and thus the page will load quicker.

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -15,9 +15,36 @@ To import all the Sass rules from GOV.UK Frontend, add the following to your Sas
...
+@import "node_modules/govuk-frontend/govuk/base";
+
+@import "node_modules/govuk-frontend/govuk/core/all";
+@import "node_modules/govuk-frontend/govuk/objects/all";
+@import "node_modules/govuk-frontend/govuk/components/footer/index";
+@import "node_modules/govuk-frontend/govuk/components/header/index";
+@import "node_modules/govuk-frontend/govuk/components/skip-link/index";
+@import "node_modules/govuk-frontend/govuk/utilities/all";
+@import 
+"node_modules/govuk-frontend/govuk/overrides/all";
```
36degrees, line 37:
```suggestion
@import "node_modules/govuk-frontend/govuk/overrides/all";
```

m-green: Oops - thanks!

source/importing-css-assets-and-javascript/index.html.md.erb
m-green: Ah thanks - updated in 148d922.

36degrees:
LGTM – needs 2i and v3.7.0 to be released?

m-green:
This has been 2i'd - now waiting for v3.7.0 release.

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -15,9 +15,37 @@ To import all the Sass rules from GOV.UK Frontend, add the following to your Sas
...
+@import "node_modules/govuk-frontend/govuk/objects/all";
+@import "node_modules/govuk-frontend/govuk/components/footer/index";
+@import "node_modules/govuk-frontend/govuk/components/header/index";
+@import "node_modules/govuk-frontend/govuk/components/skip-link/index";
+@import "node_modules/govuk-frontend/govuk/utilities/all";
+@import "node_modules/govuk-frontend/govuk/overrides/all";
+```
+You can remove lines that import parts of the CSS you do not need. 
+
+[Read more about the different parts of GOV.UK Frontend’s CSS](https://github.com/alphagov/govuk-frontend/tree/master/src/govuk).
```
hannalaakso, line 40: Non-blocking but this should really link to https://github.com/alphagov/govuk-frontend/tree/master/src/govuk#the-src-directory

PR #57, opened 2020-05-26
Add variable default values to Sass API Reference

### LINKED ISSUE #45: Include default values for settings in the Sass API reference
## What

Document the default values for settings within the Sass API reference

## Why

As a user configuring GOV.UK Frontend, knowing what the default value for a setting is helps me understand how that setting is used, and whether I need to change it.

36degrees:
Closes #45

### DIFF
--- a/source/sass-api-reference/index.html.md.erb
+++ b/source/sass-api-reference/index.html.md.erb
@@ -46,6 +46,13 @@ weight: 65
 <% end %>
 <% end %>
 
+<% if item.context.type == 'variable' %>
+##### Default value
+```scss
+$<%= item.context.name %>: <%= item.context.value %>;
+```
+<% end %>
+
 <% if item.example %>
 
 <% item.example.each do |example| %>


source/sass-api-reference/index.html.md.erb
```diff
@@ -46,6 +46,13 @@ weight: 65
 <% end %>
 <% end %>
 
+<% if item.context.type == 'variable' %>
+##### Default value
```
m-green, line 50:
```suggestion
##### Default
```

Suggest using just 'Default' for this heading, because the example code in each code has more than the value alone.

36degrees: I think I'd be inclined to keep it as is, but I can't really articulate why. Part of me just thinks 'default _what_?' I guess it is showing you what the default value is, it's just that it shows it with a bit more context to help you understand what is being shown.

Thoughts? How strongly do you feel about this one? 😁

m-green: Not massively! Feel free to keep as 'Default value'.

36degrees: Cool, OK. Let's stick with it for now and we can always revisit.

PR #67, opened 2020-07-22
Add docs on using Nunjucks macros

m-green:
As part of [Improving guidance for using Nunjucks docs](https://github.com/alphagov/govuk-design-system/issues/1281), this creates a new page in GOV.UK Frontend docs about using Nunjucks macros.

It also updates a couple of links elsewhere in the docs to point to this new page.

Thanks to @hannalaakso for her help pairing to create this new documentation.


mgladdish:
I'd say "your application and any frameworks that support it must be written in JavaScript" isn't quite right. It's worse than that because, as far as I can tell, Nunjucks can only be used within the browser or within Node. Just using Javascript isn't enough.

mgladdish:
And surely this is about using Nunjucks *templates* and macros? Not just macros alone? This reads as though you could choose to use Nunjucks templates but not the macros. Which you can't AFAIK.

source/use-nunjucks-macros/index.html.md.erb
```diff
@@ -0,0 +1,94 @@
...
+5. Go to the [default page template example](https://design-system.service.gov.uk/styles/page-template/#default) on the GOV.UK Design System website, then copy the Nunjucks code into your view file.
+
+## Importing a component
+
+You can go to any component page on the Design System website, then copy the code from the **Nunjucks** tab of any example.
+
+For example, to add the breadcrumbs component to your page, add the following code from the first example on the [breadcrumbs component page](https://design-system.service.gov.uk/components/breadcrumbs/):
+
+```javascript
+{% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
```
36degrees, line 58: Are we happy with the risk of this getting out of sync with the first example on the Breadcrumbs page?

source/use-nunjucks-macros/index.html.md.erb
```diff
@@ -0,0 +1,94 @@
...
+
+```javascript
+{{ govukButton({
+    text: "Example button text"
+}) }}
+```
+
+To see the options for a component, select the **Nunjucks** tab of the component example on any Design System website page, then select **Nunjucks macro options**.
+
+There’s a different way to [change how the page template works](https://design-system.service.gov.uk/styles/page-template/#changing-template-content).
```
36degrees, line 92: I initially read this as 'there is a second, alternative way to do this same thing' rather than 'configuring the page template works differently'.

Given this is closely related to configuring the page template, should we mention this in the 'Set up Nunjucks' as well?

source/use-nunjucks-macros/index.html.md.erb
```diff
@@ -0,0 +1,94 @@
+---
+title: Use Nunjucks macros
+weight: 45
+---
+
+# Use Nunjucks macros
```
36degrees, line 6: I wonder if this should just be called 'Use Nunjucks' as it covers two things that work in different ways:

- Extending (and configuring) the Nunjucks page template (which is not a macro)
- Using the component macros

Introducing them as two distinct things then breaking them into their own sections might help with explaining the fact that e.g. configuring the page template works differently?

source/use-nunjucks-macros/index.html.md.erb
m-green: Hmm good question. @hannalaakso - is it better to ask users to copy the code from the breadcrumbs page? I can't remember why we specifically included the code now.

source/use-nunjucks-macros/index.html.md.erb
m-green: Ah good points both. I'll solve both these issues at once by tweaking the language slightly, and moving it to the end of the 'Set up Nunjucks' section where I agree it's a better fit.

source/use-nunjucks-macros/index.html.md.erb
m-green: Good point, and I'm going to rejig the intro to this page slightly based on this and @mgladdish's second comment above.

m-green:
Thank you for the great feedback @mgladdish. I've iterated on the docs now, including updates from @36degrees's comments, so hopefully the new content is clearer.

source/use-nunjucks-macros/index.html.md.erb
hannalaakso: Yes good catch @36degrees, I think we can ask users to just copy it from the Design System.

m-green: 👍 I've updated in 0526b83

source/use-nunjucks/index.html.md.erb
```diff
@@ -0,0 +1,74 @@
...
+If you use Nunjucks, your project will be easier to update when a new version of GOV.UK Frontend is released.
+
+To use Nunjucks:
+
+- you must [install GOV.UK Frontend with Node.js package manager (npm)](/installing-with-npm/)
+- your application and any frameworks that support it must use Node.js
+
+You do not need to follow this guidance to [use Nunjucks macros in the GOV.UK Prototype Kit](https://design-system.service.gov.uk/get-started/prototyping/#using-nunjucks-macros).
+
+## Set up Nunjucks and use the page template
```
36degrees, line 21: I suspect in most cases, what the user does is going to look quite different to this, depending on what framework they're using. For example, if they're using express they might install [`express-nunjucks`](https://www.npmjs.com/package/express-nunjucks); if they're using eleventy it already has nunjucks installed and support built in.

source/use-nunjucks/index.html.md.erb
```diff
@@ -0,0 +1,74 @@
...
+    nunjucks.configure([
+      "node_modules/govuk-frontend/"
+    ], {
+      autoescape: true
+    })
+    ```
+
+    `autoescape: true` makes the output render as text by default, to help prevent cross-site scripting (XSS) attacks.
+
+4. Use the GOV.UK Frontend template by adding the following at the top of your view file:
```
36degrees, line 43: There's a step missing here which is where the configured nunjucks environment is actually hooked in to the view layer of whatever framework they're using.

Unfortunately, what that step looks like is going to depend a lot on what framework is being used.

source/use-nunjucks/index.html.md.erb
```diff
@@ -0,0 +1,74 @@
...
+
+```javascript
+{{ govukButton({
+    text: "Example button text"
+}) }}
+```
+
+To see the options for a component, select the **Nunjucks** tab of the component example on any Design System website page, then select **Nunjucks macro options**.
+
+You must sanitise any HTML you pass in to Nunjucks macros you’re using in your live application, to protect your website against cross-site scripting (XSS) attacks. Talk to your team’s security specialist about how to do this.
```
36degrees, line 74: I think [@vanitabarrett's comment on #70](https://github.com/alphagov/govuk-frontend-docs/pull/70#discussion_r463640030) applies here as well – the chances that there's a clear 'security specialist' to go and talk to for any given team seem slim to me.

source/use-nunjucks/index.html.md.erb
```diff
@@ -0,0 +1,74 @@
...
+
+    ```javascript
+    nunjucks.configure([
+      "node_modules/govuk-frontend/"
+    ], {
+      autoescape: true
+    })
+    ```
+
+    `autoescape: true` makes the output render as text by default, to help prevent cross-site scripting (XSS) attacks.
```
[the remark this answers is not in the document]
m-green: Oh thanks for pointing this out - we can probably leave this out then.

source/use-nunjucks/index.html.md.erb
```diff
@@ -0,0 +1,74 @@
...
+- you must [install GOV.UK Frontend with Node.js package manager (npm)](/installing-with-npm/)
+- your application and any frameworks that support it must use Node.js
+
+You do not need to follow this guidance to [use Nunjucks macros in the GOV.UK Prototype Kit](https://design-system.service.gov.uk/get-started/prototyping/#using-nunjucks-macros).
+
+## Set up Nunjucks and use the page template
+
+1. Make sure you installed Nunjucks when you [installed GOV.UK Frontend with npm](/installing-with-npm).
+
+2. Import the Nunjucks package at the top of your configuration file, for example:
```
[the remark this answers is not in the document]
m-green: 'Main application file' sounds good to me. We should probably document these terms...

m-green:
@36degrees @hannalaakso Updated in 250ad57 and ready for re-review! We've made it clearer that how you set up Nunjucks (generically) will depend on your environment, but our example is for users using Express.

source/use-nunjucks/index.html.md.erb
```diff
@@ -0,0 +1,71 @@
...
+To use Nunjucks:
+
+- you must [install GOV.UK Frontend with Node.js package manager (npm)](/installing-with-npm/)
+- your application and any frameworks that support it must use Node.js
+
+You do not need to follow this guidance to [use Nunjucks macros in the GOV.UK Prototype Kit](https://design-system.service.gov.uk/get-started/prototyping/#using-nunjucks-macros).
+
+## Set up Nunjucks and use the page template
+
+How you set up Nunjucks will depend on your environment. You can do the following to set up Nunjucks if you're using [Express](https://expressjs.com/).
```
36degrees, line 23: Have we tested this with a clean install of Express? I _think_ at the very least Nunjucks needs to be configured to 'install itself' against the Express app (as per https://mozilla.github.io/nunjucks/api.html#configure)

source/use-nunjucks/index.html.md.erb
```diff
@@ -0,0 +1,71 @@
...
+
+    ```javascript
+    const nunjucks = require('nunjucks')
+    ```
+
+3. Tell Nunjucks where to find the GOV.UK Frontend template and components, by adding the following to your main application file:
+
+    ```javascript
+    nunjucks.configure([
+      "node_modules/govuk-frontend/"
```
36degrees, line 37: Would this need the user's 'view' folder included as well?

m-green:
Tweaked to focus on Frontend-specific steps following a chat with Ollie on Friday.

Issue #68, opened 2020-07-23
Document how to configure Content Security Policy (CSP) to work with the inline 'js-enabled' script


hannalaakso:
## What

[GOV.UK Frontend page template](https://design-system.service.gov.uk/styles/page-template/) contains an inline script tag which adds the `js-enabled` class. [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) (CSP) can block inline scripts.

We need to tell users how to allow the script in their CSP by including the hash of the inline script tag: `sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=`. This is discussed in https://github.com/alphagov/govuk-frontend/issues/1811#issuecomment-663171834.

An example CSP could look like

```
Content-Security-Policy: default-src 'self'; script-src 'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='
```

This documentation should probably be added as a new section under [JavaScript](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#javascript). 

We should mention that: 
- we recommend teams talk to their security specialist when creating their CSP.
- we will treat any changes to the hash as a breaking change but we don’t expect the script tag to change much, if ever, so the hash should stay fairly static over time.

We could consider mentioning that teams don't need to make any changes to the script tag itself. A user who recently reviewed our workings doc on this assumed that they would also need to add an attribute to the script tag because the various ways to solve this problem (hash, nonces, SRI) all do things slightly differently and can be confused with each other; the user recommended that we confirm in our docs that no markup changes are needed.

## Why

Teams have reported this as an issue and we have no guidance to help them resolve it:
https://github.com/alphagov/govuk-frontend/issues/1811
https://github.com/alphagov/govuk-frontend/issues/1657 

## Who needs to know about this
Mark, developer

## Further detail

We have an inline script tag in the page template to prevent a flash of content for example on accordion and details components. 

Currently, if you include the page template in your application and your application implements a CSP, the inline script could be blocked from running and you might see a warning about this in the console, eg.:
 
 ```
[...]

PR #70, opened 2020-07-30
Add documentation to configure CSP to work with inline script in page template

hannalaakso:
This PR documents teams should configure their Content Security Policy to work with the inline script tag in the page template.

The reasoning for this documentation is captured in:
https://github.com/alphagov/govuk-frontend/issues/1811#issuecomment-663171834
https://github.com/alphagov/govuk-frontend-docs/issues/68

We also have a [related PR](https://github.com/alphagov/govuk-frontend/pull/1884) to test if hash in these docs matches the inline script tag in page template. 

Fixes https://github.com/alphagov/govuk-frontend-docs/issues/68

Thanks @m-green for helping to write this up 🙏

### DIFF
--- a/source/importing-css-assets-and-javascript/index.html.md.erb
+++ b/source/importing-css-assets-and-javascript/index.html.md.erb
@@ -193,3 +193,21 @@ For example:
     })
   </script>
 ```
+
+### If your JavaScript isn’t working properly
+
+If your site has a Content Security Policy (CSP), the CSP may block the inline JavaScript in the page template. You may see a warning similar to the following in the developer tools in your browser:
+
+```
+Refused to execute inline script because it violates the following Content Security Policy directive: "default-src 'self'".
+```
+
+You can unblock the JavaScript by including the following hash in your CSP:
+
+```
+sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=
+```
+
+You do not need to make any changes to the HTML.
+
+You can read guidance about [editing CSP files](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) on the MDN website.


source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -193,3 +193,19 @@ For example:
...
+
+### If your JavaScript isn’t working properly
+
+If your site has a Content Security Policy (CSP), the CSP may block the inline JavaScript in the page template. You may see a warning similar to the following in the developer tools in your browser:
+
+```
+Refused to execute inline script because it violates the following Content Security Policy directive: "default-src 'self'".
+```
+
+You can unblock the JavaScript by asking your security specialist to include the following hash in your CSP:
```
vanitabarrett, line 205: Is "security specialist" a bit too prescriptive here? Just thinking about organisations/projects where the CSP is owned by developers, this might be a bit confusing?

36degrees: +1 – I think the intention here is to make sure people who care about this are involved when the change is made? Definitely agree with that, but as it reads at the minute I think I'm not sure it'd mean anything to most people. And even if there is someone 'overseeing' the use of CSP etc, it's likely that the actual change would be made by developers.

hannalaakso: Good point 👍 

> I think the intention here is to make sure people who care about this are involved when the change is made?

Yes. The changes we usually require users to make are localised to a component or pages but changing your CSP could impact your whole app if you get it wrong. So it might be useful to ask users to take special care here and/or flag that this is something that we wouldn't necessarily expect the developer to do on their own.

What do you think? (We can always ask for help to word this in a better way)

hannalaakso: Resolved! After speaking to @m-green we've now removed the 'security specialist' mention but link to some CSP guidance.

PR #72, opened 2020-08-05
Fix style in JavaScript heading

m-green:
Fix style in heading from 'don't' to GOV.UK style 'do not'.


m-green:
Hi @hannalaakso - this was the only change to the CSP guidance from 2i.

PR #74, opened 2020-08-19
Add docs on testing with HTML fixtures

m-green:
Part of https://github.com/alphagov/govuk-frontend/issues/1830

This adds a new docs page about testing your HTML matches GOV.UK Frontend. 

Before merging we need to:

- [x] get the content pre-i'd and 2i'd
- [x] update the code example with the final code

### DIFF
--- a/source/testing-your-html/index.html.md.erb
+++ b/source/testing-your-html/index.html.md.erb
@@ -1,14 +1,14 @@
 ---
-title: Test your HTML matches GOV.UK Frontend
+title: Test if your HTML matches GOV.UK Frontend
 weight: 61
 ---
 
-# Test your HTML matches GOV.UK Frontend
+# Test if your HTML matches GOV.UK Frontend
 
 You can use our test fixtures to check you're outputting the same HTML that GOV.UK Frontend uses, when you:
 
 - create your own component macros (also called ‘partials’ or ‘templates’)
-- update GOV.UK Frontend
+- [update GOV.UK Frontend](/updating-with-npm/)
 
 ## Using the HTML test files
 
--- a/source/testing-your-html/index.html.md.erb
+++ b/source/testing-your-html/index.html.md.erb
@@ -1,0 +1,51 @@
+---
+title: Test your HTML matches GOV.UK Frontend
+weight: 61
+---
+
+# Test your HTML matches GOV.UK Frontend
+
+You can use our test fixtures to check you're outputting the same HTML that GOV.UK Frontend uses, when you:
+
+- create your own component macros (also called ‘partials’ or ‘templates’)
+- update GOV.UK Frontend
+
+## Using the HTML test files
+
+If you [installed GOV.UK Frontend with Node.js package manager (npm)](/installing-with-npm), you can find the `fixtures.json` file for each component in the `node_modules/govuk-frontend/govuk/components/COMPONENT-NAME/` folder, where `COMPONENT-NAME` is the name of the component.
+
+For example, you can find the `fixtures.json` file for the button component in the `node_modules/govuk-frontend/govuk/components/button/` folder:
+
+```js
+{
+  "component": "button",
+  "fixtures": [
+    {
+      "name": "default",
+      "options": {
+        "text": "Save and continue"
+      },
+      "html": "<button class=\"govuk-button\" data-module=\"govuk-button\">\n  Save and continue\n</button>"
+    },
+    {
+      "name": "secondary button",
+      "options": {
+        "text": "Find address",
+        "classes": "govuk-button--secondary"
+      },
+      "html": "<button class=\"govuk-button govuk-button--secondary\" data-module=\"govuk-button\">\nFind address\n</button>"
+    },
+    ...
+  ]
+}
+```
+
+Each object in the `fixtures` list is a different example of the component, where:
+
+- `name` is the name of the example
+- `options` are the options that generate this example's `html`
+- `html` is the HTML that GOV.UK Frontend outputs with these options
+
+For each example, pass `options` into your own macro and check the generated HTML matches `html`.
+
+If your HTML does not match exactly, you'll need to write your tests so they ignore known differences. For example your framework may add extra whitespace or attributes to your HTML.


source/testing-your-html/index.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+
+Each object in the list is a different example of the component, where:
+
+- `name` is the name of the example
+- `options` are the options that generate this example's `html`
+- `html` is the HTML that GOV.UK Frontend outputs with these options
+
+For each example, pass `options` into your own macro and check the generated HTML matches `html`.
+
+Your HTML may not match exactly. For example, your framework may add extra whitespace or attributes to your HTML.
```
m-green, line 50: @36degrees @vanitabarrett I had a long look at the final paragraph and realised I might be trying to over-explain things. In retrospect, it feels like enough to warn users that the HTML may not match, without going into what to do about it - as this would be fairly self-explanatory? Let me know what you think!

36degrees: I think the word 'may' makes it sound permissive? ('It's OK if the HTML doesn't match exactly')

Do we just need to be clear that they will have to do _something_…? What about something like…

> If your HTML does not match exactly – for example if your framework adds extra whitespace or attributes to your HTML – you'll need to write your tests to allow for known differences.

source/testing-your-html/index.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+]
+```
+
+Each object in the list is a different example of the component, where:
+
+- `name` is the name of the example
+- `options` are the options that generate this example's `html`
+- `html` is the HTML that GOV.UK Frontend outputs with these options
+
+For each example, pass `options` into your own macro and check the generated HTML matches `html`.
```
hannalaakso, line 48: I really liked the way we explained in pseudo code in the [proposal](https://github.com/alphagov/govuk-frontend/issues/1830#issuecomment-665075842) what we mean by checking for match, eg.

> For example, in pseudo-code:

```
fixtures = load_json_file("govuk/components/button/fixtures.json")
for fixture in fixtures
	expected_ouput = fixture.html
	actual_output = render_my_button(fixture.options)
	expect_equal(actual_output, expected_output)
end
```

Could we consider doing something similar here? I think this would really help to clarify on a high level what you need to do.

hannalaakso (commented):
Looks good 👍 Just left one comment.

We could also consider adding something the package README to document the HTML fixtures since it already [explains](https://github.com/alphagov/govuk-frontend/blob/master/package/govuk/components/accordion/README.md#component-options) what the macro options file is for. I can create another issue for that if it's something we want to do?

source/testing-your-html/index.html.md.erb
m-green: Hi Hanna - I discussed this briefly with Ollie and Vanita and we think the content works ok without it. If we did add code, I wonder whether a specific real example rather than pseudocode might work better. But I think it's worth seeing if the content works without it for now, as the written description should cover everything in the code example.

source/testing-your-html/index.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+## Using the HTML test files
+
+If you [installed GOV.UK Frontend with Node.js package manager (npm)](/installing-with-npm), you can find the `fixtures.json` file for each component in the `node_modules/govuk-frontend/govuk/components/COMPONENT-NAME/` folder, where `COMPONENT-NAME` is the name of the component.
+
+<%= warning_text('You should not test your HTML using the YAML files in the <code>src/</code> folder.') %>
+
+For example, you can find the `fixtures.json` file for the button component in the `node_modules/govuk-frontend/govuk/components/button/` folder:
+
+```js
+[
```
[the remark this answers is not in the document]
m-green: > Looks good 👍 Just left one comment.
> 
> We could also consider adding something the package README to document the HTML fixtures since it already [explains](https://github.com/alphagov/govuk-frontend/blob/master/package/govuk/components/accordion/README.md#component-options) what the macro options file is for. I can create another issue for that if it's something we want to do?

Hi @hannalaakso - ah thanks for spotting this. Are those individual READMEs in each component folder auto-generated?

source/testing-your-html/index.html.md.erb
m-green: > Looks good 👍 Just left one comment.
> 
> We could also consider adding something the package README to document the HTML fixtures since it already [explains](https://github.com/alphagov/govuk-frontend/blob/master/package/govuk/components/accordion/README.md#component-options) what the macro options file is for. I can create another issue for that if it's something we want to do?

Hi @hannalaakso - ah thanks for spotting this. Are those individual READMEs in each component folder auto-generated?

source/testing-your-html/index.html.md.erb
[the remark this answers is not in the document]
m-green: Thanks Vanita - fixed in ee2717b.

source/testing-your-html/index.html.md.erb
m-green: I've had another go in ee2717b - I've mainly used Ollie's suggestion but with slight style tweak. Let me know what you think!

source/testing-your-html/index.html.md.erb
hannalaakso: @m-green They're not autogenerated from a template but copied over from `src/govuk/components` as part of the release so each component README in `src` would need to be updated separately.

m-green: Ah ok, thanks. I wonder whether the HTML fixtures might be slightly outside the scope of these READMEs, as these READMEs seem more about 'basic' usage. Interested to hear what others think though.

36degrees: Agree – I think the section in the component READMEs is explaining that components have options and that you can see what those options are in the Design System, rather than anything to do with the `macro-options.json` file?

source/testing-your-html/index.html.md.erb
```diff
@@ -0,0 +1,53 @@
...
+You can use our test fixtures to check you're outputting the same HTML that GOV.UK Frontend uses, when you:
+
+- create your own component macros (also called ‘partials’ or ‘templates’)
+- update GOV.UK Frontend
+
+## Using the HTML test files
+
+If you [installed GOV.UK Frontend with Node.js package manager (npm)](/installing-with-npm), you can find the `fixtures.json` file for each component in the `node_modules/govuk-frontend/govuk/components/COMPONENT-NAME/` folder, where `COMPONENT-NAME` is the name of the component.
+
+<%= warning_text('You should not test your HTML using the YAML files in the <code>src/</code> folder.') %>
```
36degrees, line 17: Still not 100% sure that this is terribly important to mention – or at least that it's important enough to warrant being in a callout. If people do it nothing bad's going to happen, it's just that we might change those files without giving them any warning.

36degrees (approved):
This is looking good to me @m-green, bar not being 100% sure about the callout.

source/testing-your-html/index.html.md.erb
m-green: No probs - happy to remove this.

m-green: Removed in eaeabe9

vanitabarrett (approved):

PR #76, opened 2020-09-07
Update tech docs gem and fix indexing of error files

36degrees:
Update tech docs gem to 2.1.0.

Also fix indexing of error files – `index: false` is not a valid frontmatter option, but now we've updated to tech docs v2.1.0 we can use `prevent_indexing: true`.

This will also fix some warnings seen in the log where `index` conflicts with a built in method name:

```
WARN -- : You are setting a key that conflicts with a built-in method Middleman::Util::EnhancedHash#index defined in Hash. This can cause unexpected behavior when accessing the key as a property. You can still access the key via the #[] method.
```

Remove the index option entirely from the Google site verification file, which does not use a layout and does therefore does not output any meta tags.


hannalaakso (approved):
👍

PR #83, opened 2020-11-11
Fix broken numbering in 'Get started' content

EoinShaughnessy:
[Get started > Get the font and images working](https://frontend.design-system.service.gov.uk/get-started/#4-get-the-font-and-images-working).

In the current content, the number 1 is repeated after a bullet list.

### DIFF
--- a/source/get-started/index.html.md.erb
+++ b/source/get-started/index.html.md.erb
@@ -55,9 +55,8 @@ There are also different ways you can [import GOV.UK Frontend's CSS](/importing-
 Your component will not use the right font or images until you've added GOV.UK Frontend's assets to your application.
 
 1. Copy the following 2 folders:
-
-- `/node_modules/govuk-frontend/govuk/assets/images` folder to `<YOUR-APP>/assets/images`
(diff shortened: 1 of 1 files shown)


36degrees:
It's still not quite doing what we want for some reason. From playing around locally, it looks like it displays correctly if we remove the blank line between the first numbered bullet and the bulleted list (I don't understand why!):

```
1. Copy the following 2 folders:
  - `/node_modules/govuk-frontend/govuk/assets/images` folder to `<YOUR-APP>/assets/images`
  - `/node_modules/govuk-frontend/govuk/assets/fonts` folder to `<YOUR-APP>/assets/fonts`
```

Which then displays like this, with the correct numbering:

Screenshot 2020-11-11 at 17 33 15

If you can, I'd suggest getting the frontend docs up and running locally so you can experiment changing the markdown and see what it does.

Minor, but it'd also be good to update the commit message so that it explains what the problem is we're trying to fix. There's some [really good guidance on writing good commit messages in the GDS Way](https://gds-way.cloudapps.digital/standards/source-code.html#links-to-issue-trackers) that's worth a read.

I'd be happy to pair on this, if it'd be useful to do so.

36degrees (approved):
Good stuff! 👍🏻

PR #93, opened 2020-12-17
Fix continuous deployment

### LINKED ISSUE #91: Fix continuous deployment of Frontend Docs
## What

Fix continuous deployment of the GOV.UK Frontend docs, which is currently broken.

## Why

The deploy stage is failing on the `master` branch, which means changes are no longer being deployed automatically.

```
Starting deployment for app govuk-frontend-docs in org govuk-design-system / space frontend-docs as design-system-deploy+frontend-docs@digital.cabinet-office.gov.uk...
OK
Waiting for app to start...
Start app timeout
TIP: Application must be listening on the right port. Instead of hard coding the port, use the $PORT environment variable.
Use 'cf logs govuk-frontend-docs --recent' for more information
FAILED
Script failed with status 1
```

We think this is because we're still using the `v3` prefixed commands to deploy:

https://github.com/alphagov/govuk-frontend-docs/blob/a7f28e490627bfdcb93df7eb4764c7372433a360/bin/deploy-travis#L11-L14

And, roughly:

* v3- commands were experimental, while they worked on the API
* v7 CLI came out, and uses the v3 API in a not-experimental way (i.e. it’s supported properly)
* The v3- commands weren’t updated to work with newer versions of the API

So we think the solution is to [upgrade to the V7 version of the CLI](https://docs.cloudfoundry.org/cf-cli/v7.html), and use non-prefixed commands.

## Further details

The same issue came up in the learn-to-code repo, and was fixed by the changes in https://github.com/alphagov/learn-to-code/pull/40.

~~It might make sense to do #89 at the same time.~~

36degrees:
Install version 7 of the CloudFoundry CLI and use it to deploy the site to PaaS.

The cf push command in v7 has a new flag —strategy which [‘can deploy an app without causing downtime using —strategy rolling’][1] – which is equivalent to the `v3-zdt-push` command. We also no longer need to manually create the app or apply the manifest, which means we no longer need to extract the app path from the manifest.

These changes mirror the [changes made in alphagov/learn-to-code][2] which had the same problem.

~~**BEFORE MERGING, we need to remove 436095f which only exists to test the deployment before this gets merged to master.**~~

Closes #91 

[1]: https://docs.cloudfoundry.org/cf-cli/v7.html#:~:text=You%20can%20deploy%20an%20app%20without%20causing%20downtime%20using%20cf%20push%20app_name%20--strategy%20rolling
[2]: alphagov/learn-to-code#40

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_deploy:
   - wget -q -O - https://packages.cloudfoundry.org/debian/cli.cloudfoundry.org.key | sudo apt-key add -
   - echo "deb https://packages.cloudfoundry.org/debian stable main" | sudo tee /etc/apt/sources.list.d/cloudfoundry-cli.list
   - sudo apt-get update
-  - sudo apt-get install -y cf-cli
+  - sudo apt-get install -y cf7-cli
 
 before_script:
   - npm install
--- a/bin/deploy-travis
+++ b/bin/deploy-travis
@@ -1,14 +1,12 @@
 #!/usr/bin/env bash
 
-cf api $CF_API
-cf auth # uses CF_USERNAME / CF_PASSWORD from environment
-cf target -o $CF_ORG -s $CF_SPACE
+# Exit if the deploy script hits an error or unset variable
+set -eu
 
 # Parse app name from manifest to ensure it matches up
 APP_NAME=$(ruby -e "require 'yaml'; config = YAML.load_file('manifest.yml'); puts config['applications'][0]['name']")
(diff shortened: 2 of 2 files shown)


richardTowers (approved):
This looks correct to me.

Only nitpicks would be that you could do with `set -eu` at the top of your deploy-travis script, so it will stop if it hits an error or unset variable, and quoting `"$CF_API"`, `"$CF_ORG"` and `"$CF_SPACE"` is good practice (although they'll probably never contain whitespace, so technically it's not necessary).

AP-Hunt (approved):
From a PaaS perspective this looks spot on.

vanitabarrett (approved):

PR #98, opened 2021-01-07
Add contextual info to IE8 doc

EoinShaughnessy:
The [current IE8 content](https://frontend.design-system.service.gov.uk/supporting-ie8/#support-internet-explorer-8) doesn't specify that you only follow these steps if you've already decided to support IE8. We should call this out.

### DIFF
--- a/source/supporting-ie8/index.html.md.erb
+++ b/source/supporting-ie8/index.html.md.erb
@@ -5,9 +5,10 @@ weight: 50
 
 # Support Internet Explorer 8
 
-If you are including GOV.UK Frontend as part of your application's stylesheets
-then you will need to do some additional work to support Internet Explorer 8
-(IE8).
+You will need to do some extra work to support Internet Explorer 8 (IE8) if:
+
+- your users' browser is IE8
+- you include GOV.UK Frontend in your application’s stylesheets
 
 1. Include an HTML5 shiv.
 2. Generate an IE8-specific stylesheet.
--- a/source/supporting-ie8/index.html.md.erb
+++ b/source/supporting-ie8/index.html.md.erb
@@ -5,10 +5,7 @@ weight: 50
 
 # Support Internet Explorer 8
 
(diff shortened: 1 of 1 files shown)


source/supporting-ie8/index.html.md.erb
```diff
@@ -5,9 +5,10 @@ weight: 50
 
 # Support Internet Explorer 8
 
-If you are including GOV.UK Frontend as part of your application's stylesheets
-then you will need to do some additional work to support Internet Explorer 8
-(IE8).
+You will need to do some extra work to support Internet Explorer 8 (IE8) if:
+
+- your users' browser is IE8
+- you include GOV.UK Frontend in your application’s stylesheets
```
36degrees, line 11: Whether or not a service needs to support IE8 is a decision that the service team really need to make based on their own analytics and research.

Can we frame this to tell service teams who have _already decided_ they need to support IE8 understand what they need to do?

I am also not sure we need the clause about 'If you are including GOV.UK Frontend as part of your application's stylesheets' – if they're using GOV.UK Frontend then they absolutely should be including the stylesheets!

With that in mind, what if we simplified this to something like:

```suggestion
If your service needs to support Internet Explorer 8 (IE8) then you'll need to follow these additional steps:
```

EoinShaughnessy:
```suggestion
If your service needs to support Internet Explorer 8 (IE8), then follow these extra steps:
```

EoinShaughnessy: @36degrees How's that^?

36degrees: Looks good to me 👍🏻

cmcnallygds (approved):
LGTM

SteveEvansGDS (approved):

EoinShaughnessy:
Update has passed 2i. Will merge PR.

Issue #104, opened 2021-03-12
Fix accessibility issues by updating the govuk_tech_docs gem to 2.2.0 [before 31 March]


36degrees:
## What

First, update the govuk_tech_docs gem from 2.1.0 to 2.2.0.

<details>
<summary>Change log</summary> 

> ## 2.2.0
> 
> ### Accessibility Fixes
> 
> - [#205: Improve table of contents accessibility](https://github.com/alphagov/tech-docs-gem/pull/205)
> - [#209: Some search and keyboard navigation updates](https://github.com/alphagov/tech-docs-gem/pull/209)
> - [#210: Stop linking images to new tabs](https://github.com/alphagov/tech-docs-gem/pull/210)
> - [#214: Implement row level table headings to allow accessible tables with row headings](https://github.com/alphagov/tech-docs-gem/pull/214)
> - [#215: Add tabindex and focus states to code blocks](https://github.com/alphagov/tech-docs-gem/pull/215)
> 
> ### Docs
> 
> - [#206: Remove tdt docs content from readme file](https://github.com/alphagov/tech-docs-gem/pull/206)
> 
> ## 2.1.1
> 
> ### Fixes
> 
> We’ve made fixes in the following pull requests:
> 
> - [#199: Fix page expiry box link colours (hover and focus) ](https://github.com/alphagov/tech-docs-gem/pull/199)
> 
> ### Ruby version bump
> 
> We've updated the Ruby version supported:
> 
> - [#201: Bump ruby to 2.7.2](https://github.com/alphagov/tech-docs-gem/pull/201)
</details>

Then identify and fix any tables which require row level headers.

There is a single table in [migrating from legacy products](https://frontend.design-system.service.gov.uk/migrating-from-legacy-products/#1-uninstall), and multiple tables in the Sass API reference docs.

## Why

Version 2.2.0 of the tech docs gem is now live and incorporates fixes to a number of accessibility issues in the technical documentation template which are referenced in our accessibility statement:

- Some tables in content do not have column row headers when needed.
- Users cannot tab to code examples to scroll them using the keyboard.
- The search field and the ‘Hide table of contents’ button are not placed logically in the focus order.
- Users can focus elements that are hidden visually behind other content, for example when using the search field.
- Images link to a new tab without telling users, and do not have link text.

## Done when

- [x] Gem updated
- [ ] Tables include row headers where required
[...]

PR #106, opened 2021-03-17
Update govuk_tech_docs from 2.1.0 to 2.2.0

EoinShaughnessy:
Partially addresses https://github.com/alphagov/govuk-frontend-docs/issues/104


hannalaakso:
(Reviewed following the approach that @36degrees took in https://github.com/alphagov/prototype-kit-training/pull/54#issuecomment-800266427)

## Resolved accessibility issues

Verified that this update fixes the [accessibility issues listed in the Tech Docs Template accessibility statement](https://tdt-documentation.london.cloudapps.digital/accessibility/#using-the-technical-documentation-template-for-your-own-documentation), as follows:

### Some tables in content do not have table row headers when needed

We’ll address this issue in a separate PR.

### Some code examples use a colour that has very low contrast with the background colour

I can’t see any references to `#b26749` in `manifest.css` but `#d4351c` is referenced. 

### Users cannot tab to code examples to scroll them using the keyboard

Verified as fixed.

### The search field and the ‘Hide table of contents’ button are not placed logically in the focus order

Verified as fixed.

### Users can focus elements that are hidden visually behind other content, for example when using the search field

Verified as fixed.

### Images link to a new tab without telling users, and do not have link text

There are no images used in the Frontend docs.

### Some subheadings have links that contain no text

As explained in the guidance doc, this is not an issue like we originally thought.

### The default page header has duplicate links to the service page

There are no additional links included in the header.

### When a page has a review banner, there are duplicate links in the visible and hidden version of the banner

Review banners are not used in the Frontend docs.

### The search field does not have a submit button

Verified as fixed.

hannalaakso (approved):

Issue #108, opened 2021-03-25
Fix accessibility issues by updating the govuk_tech_docs gem to 2.2.2 [before 31 March]


EoinShaughnessy:
## What

Update the govuk_tech_docs gem from 2.2.1 to 2.2.2.

Then check if the update has fixed the hidden-content issue described in the [guidance for accessibility-fixes](https://docs.google.com/document/d/1QlTbEV6xyUsc5Tb94Iw5GrnifCuzaZB8TDIKmV5QDCs/edit#heading=h.w7nppph95a2s).

## Why

Version 2.2.2 of the Tech Docs Template fixes some outstanding issues with version 2.2.1. For example, users being able, when they use the search field, to focus elements that are 'hidden' behind other content.

## Done when

- [ ] Gem updated
- [ ] Verified that users cannot focus hidden elements by using the search field

PR #109, opened 2021-03-26
Update Tech Docs Gem to fix accessibility issues

EoinShaughnessy:
Fixes [#108](https://github.com/alphagov/govuk-frontend-docs/issues/108) (Fix accessibility issues by updating the govuk_tech_docs gem to 2.2.2 [before 31 March]).

Two days ago (24 March 2021), we [updated our Tech Docs Gem to incorporate several accessibility fixes in the Technical Documentation Template](https://github.com/alphagov/govuk-frontend-docs/issues/104).

Since then, we have learned that a new version of the Template has been released to fix some outstanding issues with the previous version. For example, users being able, when they use the search field, to focus elements that are 'hidden' behind other content.

This PR will update our Tech Docs Gem to the latest version of the Template. We can then test whether users can focus elements that should remain hidden to them. Point 5 of the [accessibility-fixes guidance](https://docs.google.com/document/d/1QlTbEV6xyUsc5Tb94Iw5GrnifCuzaZB8TDIKmV5QDCs/edit#heading=h.r8c1hz882gha) describes how to test for this behaviour.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -45,7 +45,7 @@ GEM
     fast_blank (1.0.0)
     fastimage (2.2.3)
     ffi (1.15.0)
-    govuk_tech_docs (2.2.1)
+    govuk_tech_docs (2.2.2)
       chronic (~> 0.10.2)
       middleman (~> 4.0)
       middleman-autoprefixer (~> 2.10.0)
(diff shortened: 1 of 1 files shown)


36degrees (approved):
Thanks @EoinShaughnessy 👍🏻  It looks like this does fix the bug as described in https://github.com/alphagov/tech-docs-gem/issues/221, and point 5 of the checklist.

PR #110, opened 2021-03-26
Add Github issue templates

vanitabarrett:
Closes https://github.com/alphagov/govuk-frontend-docs/issues/105

Adds the following Github issue templates to the frontend docs repo:
- bug report
- feature request
- documentation
- internal card
- link to getting support

These issue templates match the templates already added to the govuk-design-system repo.

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation.md
+++ b/.github/ISSUE_TEMPLATE/documentation.md
@@ -1,0 +1,25 @@
+---
+name: "📖 Documentation"
+about: Add new documentation, or report missing, incorrect or unclear documentation
+title: ''
+labels: "documentation, awaiting triage"
+assignees: ''
+
+---
+
+<!--
+    Please fill in as much of the template below as you’re able to. If you're unsure whether the issue already exists or how to fill in the template, open an issue anyway. Our team will help you to complete the rest.
+
+    Your issue might already exist. If so, add a comment to the existing issue instead of creating a new one. You can find existing issues here: https://github.com/alphagov/govuk-frontend-docs/issues
+
+    We also welcome pull requests from users, if you feel comfortable doing so.
+-->
+
+## Related documentation
+<!-- Does this issue refer to a gap or mistake in some existing documentation? Provide a link if possible. -->
+
+## Suggestion
+<!-- How could this documentation be improved? -->
+
+## Evidence (where applicable)
+<!-- Please provide any relevant user research or evidence to support this change. -->
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,30 @@
+---
+name: "🐛 Bug report"
+about: Report a bug or regression
+title: ''
+labels: "\U0001F41B bug, awaiting triage"
+assignees: ''
+
+---
+
+<!--
+    Please fill in as much of the template below as you’re able to. If you're unsure whether the issue already exists or how to fill in the template, open an issue anyway. Our team will help you to complete the rest.
+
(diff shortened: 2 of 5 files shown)


.github/ISSUE_TEMPLATE/documentation.md
```diff
@@ -0,0 +1,25 @@
+---
```
hannalaakso, line 1: @EoinShaughnessy Just flagging this particular one for your eyes. I know it's the same one we use in the Design System repo, just thought you'd like to have a look since this repo is more documentation focused and we might possibly want to do something different here.

hannalaakso (approved):
Thanks @vanitabarrett, looks good to me 👍

.github/ISSUE_TEMPLATE/documentation.md
```diff
@@ -0,0 +1,25 @@
...
+<!--
+    Please fill in as much of the template below as you’re able to. If you're unsure whether the issue already exists or how to fill in the template, open an issue anyway. Our team will help you to complete the rest.
+
+    Your issue might already exist. If so, add a comment to the existing issue instead of creating a new one. You can find existing issues here: https://github.com/alphagov/govuk-frontend-docs/issues
+
+    We also welcome pull requests from users, if you feel comfortable doing so.
+-->
+
+## Related documentation
+<!-- Does this issue refer to a gap or mistake in some existing documentation? Provide a link if possible. -->
```
EoinShaughnessy, line 19: By "gap," do we mean some entirely new content (like a full page) the user wants us to create, or a gap in existing docs?

If the former, then should we tweak this text to also cover requests for new content?

First go at alternative wording:

"Does this issue refer to existing documentation? If so, please provide a link to the documentation. Alternatively, if this issue requests new content, tell us what you'd like to see."

.github/ISSUE_TEMPLATE/documentation.md
```diff
@@ -0,0 +1,25 @@
...
+
+<!--
+    Please fill in as much of the template below as you’re able to. If you're unsure whether the issue already exists or how to fill in the template, open an issue anyway. Our team will help you to complete the rest.
+
+    Your issue might already exist. If so, add a comment to the existing issue instead of creating a new one. You can find existing issues here: https://github.com/alphagov/govuk-frontend-docs/issues
+
+    We also welcome pull requests from users, if you feel comfortable doing so.
+-->
+
+## Related documentation
```
EoinShaughnessy, line 18: What if there's no related documentation (for example, if a user requests a new page instead of a change to existing content)?

.github/ISSUE_TEMPLATE/documentation.md
vanitabarrett: I think it could refer to both - a missing section; an edge case that perhaps isn't covered or a piece of functionality that isn't documented at all.

The description of this template says "Add new documentation, or report missing, incorrect or unclear documentation" so it should cover both scenarios.

.github/ISSUE_TEMPLATE/documentation.md
vanitabarrett: Yeah, good point. This template is meant to cover that scenario, but it does lean quite heavily towards assuming the user is suggesting a mistake/gap rather than new documentation. We do suggest that users fill out as much of the template as possible and imply the sections are optional, but I think there's probably more we could do... Welcome to any suggestions!

EoinShaughnessy: Ah I see - thanks. The template as it is now is probably fine then. (Even if users request a wholly new page, it'll presumably relate to an existing section somehow.) We can always make changes later based on feedback.

EoinShaughnessy (approved):

PR #129, opened 2021-06-04
Tweak reference to CSP

36degrees:
Content Security Policy is defined using headers or meta tags rather than files. As it's not defined in a file you can edit, 'editing CSP files' doesn't really make sense here. The page we're linking to is a pretty general broad guide on CSP, that touches on a lot more than just how to edit it, so just refer to it as general guidance on CSP.

Also make the destination (the MDN website) part of the link text, so that someone navigating out of context still understands that it takes them to an external page.


EoinShaughnessy (approved):
Looks good!

PR #139, opened 2021-07-28
Update to GOV.UK Tech Docs v2.4.1

36degrees:


36degrees:
Link checker has picked up on a broken reference for the fallback PNG in the header, which comes from the gem.

Have raised a PR here: https://github.com/alphagov/tech-docs-gem/pull/248

lfdebrux (approved):
Looks to be working locally and on Netlify, approving 🥇

PR #144, opened 2021-09-14
Rotate CF_PASSWORD

lfdebrux:
Part of https://github.com/alphagov/design-system-team-internal/issues/511


hannalaakso (approved):
As just discussed with @lfdebrux, the easiest way to review this is to merge it and check that Travis still runs correctly🤞

Issue #89, opened 2020-12-01
Move CI/CD to GitHub Actions


36degrees:
## What

Move Frontend Docs' CI/CD from Travis to GitHub Actions.

## Why

Travis' changes to our plan means that builds are getting stuck in a queue, which means they can take around 30 minutes to run.

Given that the direction of travel for the organisation seems to be towards GitHub Actions, and to be consistent with the approach we're taking with the Prototype Kit, move to GitHub Actions.

## Who needs to know about this
Hanna, Laurence, Eoin 

## Further detail

Travis currently (as per [the config](https://github.com/alphagov/govuk-frontend-docs/blob/4197a615714525f060ee795ade258db3b4b832fe/.travis.yml)) does two things:

- Builds the site and runs a link checker
- On the master branch, if the site builds and link checker passes, deploys the site to PaaS

Equivalent jobs need to be run using GitHub Actions.

There's an example of a workflow that deploys to PaaS from National Lottery Heritage Fund here:
https://github.com/heritagefund/funding-frontend/blob/master/.github/workflows/deploy-to-production-workflow.yml

We need to make sure that the production site can still be indexed, as [the check is currently based on the `TRAVIS_BRANCH` environment variable](https://github.com/alphagov/govuk-frontend-docs/blob/4197a615714525f060ee795ade258db3b4b832fe/config.rb#L8-L9).

We should make sure that the secrets are stored at the environment level with branch protection, as per https://github.com/alphagov/design-system-team-internal/issues/446.

## Done when

- [ ] Use of secrets signed off with IA
- [ ] Existing open PRs merged / closed where possible to avoid needing to rebase
- [x] GitHub Actions config added
- [x] GitHub secrets added at environment level with branch protection
- [x] Travis config removed
- [x] `noindex` meta tag condition updated
- [ ] Required status checks updated
- [x] Updated CODEOWNERS file
- [x] Documentation updated:

https://github.com/alphagov/govuk-frontend-docs/blob/655c6e290463fcfab6d2be20210f8cbd7d9620d9/.rubocop.yml#L5

https://github.com/alphagov/govuk-frontend-docs/blob/d62c8d3164ba7fd957e4e06bb22ef6ef3ec061e3/config/tech-docs.yml#L35-L36

PR #145, opened 2021-09-16
Add CODEOWNERS file to protect secrets

lfdebrux:
Part of https://github.com/alphagov/design-system-team-internal/issues/455

We need to make sure only staff can access secrets and/or deploy to production.

This commit adds a [CODEOWNERS file] to ensure that any changes to our GitHub Actions or Travis CI pipelines are reviewed by a developer, to avoid the (unlikely) scenario where a PR that changes our CI/CD is approved by someone who isn't a developer and doesn't spot the significance of the change.

Note that this file should be updated when we migrate away from Travis (https://github.com/alphagov/govuk-frontend-docs/issues/89).

[CODEOWNERS file]: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,4 @@
+# CI/CD (which has access to secrets and to production)
+.github/workflows/  @alphagov/design-system-developers
+.travis/     @alphagov/design-system-developers
+.travis.yml  @alphagov/design-system-developers


lfdebrux:
> @lfdebrux Do we need to add anything in the `/deploy` directory to this?

Hmm, I think that's outside the scope of this ticket... as far as I can tell the deploy directory isn't involved in the actual deploy pipeline or touch secrets at all, it just has the nginx configuration for the PaaS app.

I guess we could have a conversation about whether we want to extend code owners protection to other parts of our repos?

vanitabarrett:
Thanks @lfdebrux , happy to merge this as-is and we can revisit extending codeowners protection as a separate piece of work

vanitabarrett (approved):

PR #146, opened 2021-09-20
Update quotes in code example

36degrees:
The example of what the dependency might look like in a user's `package.json` uses a mix of single and typographic quotes.

Only double quotes are supported in JSON, so update the example to use them instead.


EoinShaughnessy (approved):
Looks good!

36degrees:
@EoinShaughnessy thanks for the speedy review! ⚡

PR #151, opened 2021-10-05
Document silencing deprecation warnings in Dart Sass

36degrees:
We can't currently make the changes in GOV.UK Frontend to resolve the deprecation warnings without also removing support for compilers other than Dart Sass.

However, since Dart Sass 1.35 it's possible to silence deprecation warnings coming from dependencies.

Document how to do this, pointing to the relevant parts of the Sass docs for more information.

See https://github.com/alphagov/govuk-frontend/issues/2238 for more details.


36degrees:
@EoinShaughnessy hope this is an alright first attempt. Let me know if it'd be useful to talk through any of this.

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -72,6 +72,23 @@ If you want to override GOV.UK Frontend's styles with your own styles, `@import`
 
 If your project uses GOV.UK Frontend toolkit, GOV.UK Template or GOV.UK Elements, you can [configure GOV.UK Frontend to work with them](/compatibility-mode/).
 
+### Silence deprecation warnings in Dart Sass
+
+If you are using Dart Sass v1.33.0 or greater, you may see deprecation warnings when compiling your Sass:
```
EoinShaughnessy, line 77: Looked up Dart Sass releases, and they omit 'v' or 'version', so I'm suggesting this:

```suggestion
If you're using Dart Sass 1.33.0 or greater, you may see deprecation warnings when compiling your Sass:
```

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -72,6 +72,23 @@ If you want to override GOV.UK Frontend's styles with your own styles, `@import`
...
+
+```
+DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
+```
+
+We are currently unable to fix these deprecation warnings without breaking support for LibSass.
```
EoinShaughnessy, line 83:
```suggestion
We're currently unable to fix these deprecation warnings without breaking support for LibSass.
```

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -72,6 +72,23 @@ If you want to override GOV.UK Frontend's styles with your own styles, `@import`
...
+
+You can however silence the warnings. Ensure you are using Dart Sass 1.35.0 or greater and then:
```
EoinShaughnessy, line 85:
```suggestion
However, you can silence the warnings caused by dependencies, including GOV.UK Frontend. Make sure you're using Dart Sass 1.35.0 or greater, and then:
```

Would also move this content back into the previous paragraph, to make it the next sentence after '..without breaking support for LibSass'.

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -72,6 +72,23 @@ If you want to override GOV.UK Frontend's styles with your own styles, `@import`
...
+
+- if you are calling the sass compiler from the command line, pass [the `--quiet-deps` flag](https://sass-lang.com/documentation/cli/dart-sass#quiet-deps)
```
EoinShaughnessy, line 87:
```suggestion
- if you're calling the Sass compiler from the command line, [pass the `--quiet-deps` flag](https://sass-lang.com/documentation/cli/dart-sass#quiet-deps)
```

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -72,6 +72,23 @@ If you want to override GOV.UK Frontend's styles with your own styles, `@import`
...
+- if you are using the JavaScript API, include [`quietDeps: true`](https://sass-lang.com/documentation/js-api#quietdeps) in the `options` object
```
EoinShaughnessy, line 88:
```suggestion
- if you're using the JavaScript API, [include `quietDeps: true`](https://sass-lang.com/documentation/js-api#quietdeps) in the `options` object
```

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -72,6 +72,23 @@ If you want to override GOV.UK Frontend's styles with your own styles, `@import`
...
+
+This will only silence deprecation warnings that are caused by dependencies, including GOV.UK Frontend. You will still see deprecation warnings if you are using `/` for division in your own Sass code.
```
EoinShaughnessy, line 90: Suggesting we cut the dependencies-related sentence from here, if we're now mentioning it in line 85 instead:

```suggestion
You will still see deprecation warnings if you're using `/` for division in your own Sass code.
```

source/importing-css-assets-and-javascript/index.html.md.erb
EoinShaughnessy: Also, 'greater' or 'higher'?

source/installing-with-npm/index.html.md.erb
```diff
@@ -19,6 +19,8 @@ weight: 10
 
 3. Install [Dart Sass](https://www.npmjs.com/package/sass) - version 1.0.0 or higher.
 
+    If you are using Dart Sass v1.33.0 or greater, you may see deprecation warnings when compiling your Sass. You can [silence these deprecation warnings](/importing-css-assets-and-javascript/#silence-deprecation-warnings-in-dart-sass) if you need to.
```
EoinShaughnessy, line 22:
```suggestion
    If you're using Dart Sass 1.33.0 or greater, you may see deprecation warnings when compiling your Sass. You can [silence these deprecation warnings](/importing-css-assets-and-javascript/#silence-deprecation-warnings-in-dart-sass) if you need to.
```

Do we need call out that it silences warnings caused by dependencies?

Also, is it a 'greater' or a 'higher' version?

EoinShaughnessy (commented):
@36degrees Content looks good! Just added a few comments. Give me a shout when you're happy for it to go for 2i!

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -72,6 +72,23 @@ If you want to override GOV.UK Frontend's styles with your own styles, `@import`
 
 If your project uses GOV.UK Frontend toolkit, GOV.UK Template or GOV.UK Elements, you can [configure GOV.UK Frontend to work with them](/compatibility-mode/).
 
+### Silence deprecation warnings in Dart Sass
```
EoinShaughnessy, line 75: Is 'silence' the most direct way of saying what it is users can do here? What about 'turn off'?

Also, does this title suggest users can silence all deprecation warnings (whereas line 90 says it's only warnings caused by dependencies)?

source/importing-css-assets-and-javascript/index.html.md.erb
36degrees: I think they get used interchangeably, but I'd use 'greater' on the basis that version constraints often use the '>' (greater than) symbol, and [the node-semver documentation also uses 'greater than'](https://github.com/npm/node-semver#ranges).

Greater is perhaps a bit more formal, but it also seems more accurate as it's rooted in _value_ whereas higher is rooted in _height_?

Thoughts?

source/importing-css-assets-and-javascript/index.html.md.erb
36degrees: > Is 'silence' the most direct way of saying what it is users can do here? What about 'turn off'?

It matches the terminology used in the Sass docs here:

> This is useful for silencing deprecation warnings that you can’t fix on your own.
>
> – https://sass-lang.com/documentation/js-api#quietdeps

I think it's a verb that users would tend to associate with warnings / errors / logs, so I think it helps to front-load the heading so users who are looking for it would recognise it quickly. It also seems slightly more accurate than turn off – the warnings are still 'happening', it's just you've opted to ignore them.

> Also, does this title suggest users can silence all deprecation warnings (whereas line 90 says it's only warnings caused by dependencies)?

Will update to 'Silence deprecation warnings from dependencies in Dart Sass'?

source/installing-with-npm/index.html.md.erb
36degrees: > Why would users need to - because the warnings are an unwelcome distraction?

Pretty much, yes. They're noise, as they're not able to do anything about them, and they may get mixed up with warnings about code that is in their control that might otherwise be missed.

source/installing-with-npm/index.html.md.erb
```diff
@@ -19,6 +19,8 @@ weight: 10
 
 3. Install [Dart Sass](https://www.npmjs.com/package/sass) - version 1.0.0 or higher.
 
+    If you're using Dart Sass 1.33.0 or greater, you may see deprecation warnings when compiling your Sass. You can [silence these deprecation warnings](/importing-css-assets-and-javascript/#silence-deprecation-warnings-in-dart-sass) if you need to.
```
EoinShaughnessy, line 22: @36degrees What about:

```suggestion
    If you're using Dart Sass 1.33.0 or greater, you may see deprecation warnings when compiling your Sass. [You can silence the warnings if dependencies have caused them.](/importing-css-assets-and-javascript/#silence-deprecation-warnings-in-dart-sass)
```

Issue #153, opened 2021-10-13
Strengthen recommendations on Sass implementations in install guide


vanitabarrett:
## Related documentation
Our current documentation suggests users install Dart Sass, but we also reference Ruby Sass and Lib Sass:

> Install Dart Sass - version 1.0.0 or higher.
> GOV.UK Frontend also works with:

> LibSass - version 3.3.0 or higher
> Ruby Sass - version 3.4.0 or higher

> However, we recommend you do not use either LibSass or Ruby Sass, as they’ve been deprecated.

Given that both Ruby Sass and LibSass are deprecated and unsupported (they won't be getting new features), we might want to consider strengthening these recommendations. 

## Suggestion

We might be forced to stop supporting LibSass and Ruby Sass in the near future, when Dart Sass 2 is released. Should we remove them from the installation guide, to warn users? Perhaps something like:

> Install Dart Sass - version 1.0.0 or higher.
> We recommend you do not use either LibSass or Ruby Sass, as they've been deprecated and won't receive any updates. GOV.UK Frontend currently supports LibSass version 3.3.0+ and Ruby Sass 3.4.0+, but this support may be dropped in future.

PR #154, opened 2021-10-13
Add info about Sass deprecations to install guide

EoinShaughnessy:
Fixes [#153](https://github.com/alphagov/govuk-frontend-docs/issues/153).

This PR adds content to the [npm installation requirements](https://frontend.design-system.service.gov.uk/installing-with-npm/#install-with-node-js-package-manager-npm).

We're asking users not to use either LibSass or Ruby Sass, as we may stop supporting these implementations in future.


EoinShaughnessy:
Content has passed pre-i, so I've sent it for 2i.

EoinShaughnessy:
Content has passed 2i.

PR #158, opened 2021-11-04
Test build job for GitHub Actions

hannalaakso:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+on: [push, pull_request, workflow_dispatch]
+
+# Separate build and deploy jobs so we can build all branches and pull requests
+# but only deploy from main
+jobs:
+  build:
+    name: Build & Test
+    runs-on: ubuntu-latest
+
```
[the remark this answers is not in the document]
hannalaakso: Looks like it's used in the Design System to determine whether to allow or indexing or not [here](https://github.com/alphagov/govuk-design-system/blob/b79f7858e3e1ca91cc3c7732664b3542ce964abd/views/layouts/_generic.njk#L9-L11) - we cover this in a different way in https://github.com/alphagov/govuk-frontend-docs/pull/158/commits/36d3351c100480f7a73b5e4271244799e56c0c39 so we don't need to set the environment here.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,108 @@
...
+        run: |
+          echo "Logging into $CF_ORG/$CF_SPACE..."
+          cf api "${CF_API}"
+          cf auth
+          cf target -o "${CF_ORG}" -s "${CF_SPACE}"
+      # Parse app name from manifest to ensure it matches up
+      - name: Fetch app name from manifest
+        run: echo "APP_NAME=$(ruby -e "require 'yaml'; config = YAML.load_file('manifest.yml'); puts config['applications'][0]['name']")" >> $GITHUB_ENV
+
+      - name: Cancel any existing running deploys to avoid conflict
```
vanitabarrett, line 98: Do we want to do this using `concurrency` instead, like in the Design System refactoring PR?

lfdebrux: I thought it would be best to wait for review of https://github.com/alphagov/govuk-design-system/pull/1971 to finish before incorporating changes into other repos, just in case they were changes needed.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,108 @@
...
+        run: bundle exec rake
+
+      - name: Check links
+        run: npm run check-links
+
+  # Deploy Frontend Docs to production when the main branch is changed
+  # Github Actions is not involved in deploying PR or branch previews – these are handled by Netlify
+  # Only runs if the production environment protection rules are fulfilled
+  deploy:
+    environment: production
```
vanitabarrett, line 40: @lfdebrux @hannalaakso What is this being used for?

lfdebrux: The environment provides an extra level of security for secrets, see https://github.com/alphagov/design-system-team-internal/issues/446. It also is the thing that causes GitHub to create a deployment notification on Slack.

vanitabarrett (approved):

PR #161, opened 2021-11-11
Add task to GitHub Actions to upload artifact

hannalaakso:
Add upload artifact task to deploy job to share data between the build and deploy jobs so we don't need to run `npm run build` again on deploy. 

This will upload the deploy folder as an artifact so it can be downloaded and used in the deploy job.

This should allow the currently failing deployment job for for https://github.com/alphagov/govuk-frontend-docs/pull/158 to complete.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,15 @@ jobs:
       - name: Check links
         run: npm run check-links
 
+      # Share data between the build and deploy jobs so we don't need to run `bundle exec middleman build` again on deploy
+      # Upload the deploy folder as an artifact so it can be downloaded and used in the deploy job
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):

.github/workflows/ci.yaml
```diff
@@ -33,6 +33,15 @@ jobs:
       - name: Check links
         run: npm run check-links
 
+      # Share data between the build and deploy jobs so we don't need to run `npm run build` again on deploy
```
vanitabarrett, line 36: Do we mean `bundle exec middleman build` rather than `npm run build`?

vanitabarrett (approved):
Just a small comment about clarifying the code comment

.github/workflows/ci.yaml
hannalaakso: Good catch 👍

PR #164, opened 2021-11-12
Refactor GitHub Actions workflows and add concurrency limit to deploys

lfdebrux:
Split the `ci.yaml` workflow into two: `test.yaml` for pull request checks, and `deploy.yaml` for after merge to main. To reduce code duplication the deploy workflow calls the test workflow using the beta reusable workflow feature [[1]].

[1]: https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

The advantage of splitting the workflows like this is that we can then give the entire deploy workflow a concurrency limit [[2]], without affecting test workflows. Having a concurrency limit is an alternative to the step which cancels any in-progress
deploys, which would have resulted in failed deployment messages.

---

This is the equivalent of [PR #1971 in the govuk-design-system repo](https://github.com/alphagov/govuk-design-system/pull/1971).

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,46 +1,14 @@
-name: CI Workflow
+name: Deploy
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
 
-# Separate build and deploy jobs so we can build all branches and pull requests
-# but only deploy from main
 jobs:
   build:
-    name: Build & Test
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v2
-
-      - uses: actions/setup-node@v2
-        with:
-          node-version: '14'
-
-      - uses: ruby/setup-ruby@v1
-        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yaml
```diff
@@ -99,6 +67,7 @@ jobs:
           cf api "${CF_API}"
           cf auth
           cf target -o "${CF_ORG}" -s "${CF_SPACE}"
+
       # Parse app name from manifest to ensure it matches up
       - name: Fetch app name from manifest
         run: echo "APP_NAME=$(ruby -e "require 'yaml'; config = YAML.load_file('manifest.yml'); puts config['applications'][0]['name']")" >> $GITHUB_ENV
```
vanitabarrett, line 77: Hm, does this need updating now we've moved the `manifest.yml` file? Although it seems to be working at the moment 🤔

lfdebrux: Because the manifest file will end up in the root of the working directory either way, it should be fine.

vanitabarrett (approved):

PR #169, opened 2022-01-24
Remove beta tag

36degrees:
The GOV.UK Design System, which includes GOV.UK Frontend, [passed its live assessment on 16 November 2021][1].

Remove the beta tag from the header, as this is now a live service.

[1]: https://www.gov.uk/service-standard-reports/gov-uk-design-system-live-reassessment


lfdebrux (approved):
🎉

PR #177, opened 2022-04-07
Add documentation for importing GOV.UK Frontend ES modules

vanitabarrett:
## What
Adds documentation for how to install GOV.UK Frontend ES modules (new functionality, added in https://github.com/alphagov/govuk-frontend/pull/2586) 

## Why
We need to tell people how to use the new ES modules if they want to. We should also use this is an opportunity to recommend people only install the component JavaScript that their application actually needs, as it has a performance benefit in comparison to loading everything regardless of whether it's being used or not.

**Note: this should only be merged once https://github.com/alphagov/govuk-frontend/pull/2586 has been released in a new version of GOV.UK Frontend**

### DIFF
--- a/source/importing-css-assets-and-javascript/index.html.md.erb
+++ b/source/importing-css-assets-and-javascript/index.html.md.erb
@@ -169,16 +169,18 @@ Then import the JavaScript file before the closing `</body>` tag of your HTML pa
 If you decide to import using a bundler, we recommend you use `import` to only import the JavaScript for components you're using in your service. For example:
 
 ```javascript
-import { SkipLink, Header } from 'govuk-frontend'
+import { SkipLink, Radios } from 'govuk-frontend'
 
 var $skipLink = document.querySelector('[data-module="govuk-skip-link"]')
 if ($skipLink) {
   new SkipLink($skipLink).init()
 }
 
-var $header = document.querySelector('[data-module="govuk-header"]')
-if ($header) {
-  new Header($header).init()
+var $radios = document.querySelectorAll('[data-module="govuk-radios]')
+if ($radios) {
+  for (var i = 0; i < $radios.length; i++) {
+    new Radios($radios[i]).init()
+  }
 }
 ```
 
--- a/source/importing-css-assets-and-javascript/index.html.md.erb
+++ b/source/importing-css-assets-and-javascript/index.html.md.erb
@@ -164,6 +164,20 @@ Then import the JavaScript file before the closing `</body>` tag of your HTML pa
 </body>
 ```
 
+#### Select and initialise an individual component
+
+You can select and initialise a specific component by using its `data-module` attribute. For example, use `govuk-radios` to initialise the first radio component on a page:
+
+```html
+  <script>
+    var Radios = window.GOVUKFrontend.Radios
+    var $radio = document.querySelector('[data-module="govuk-radios"]')
+    if ($radio) {
+      new Radios($radio).init()
+    }
+  </script>
+```
+
 ### Import JavaScript using a bundler
 
 If you decide to import using a bundler, we recommend you use `import` to only import the JavaScript for components you're using in your service. For example:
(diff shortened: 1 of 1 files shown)


source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -166,13 +166,50 @@ Then import the JavaScript file before the closing `</body>` tag of your HTML pa
...
+```
+
+If you need to import all of GOV.UK Frontend's components, then run the `initAll` function to initialise them:
 
 ```javascript
 import { initAll } from 'govuk-frontend'
 initAll()
 ```
 
+If you're using Webpack, you may need to add the following to your Webpack config file:
```
vanitabarrett, line 194: Less sure about this section, as we might need to keep an eye on it and maintain it. But it seems pretty necessary for users using Webpack, so we might just have to take on that burden for a bit.

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -164,36 +164,73 @@ Then import the JavaScript file before the closing `</body>` tag of your HTML pa
...
+```html
+  <script>
+    var Radios = window.GOVUKFrontend.Radios
+    var $radio = document.querySelector('[data-module="govuk-radios"]')
+    if ($radio) {
+      new Radios($radio).init()
+    }
+  </script>
+```
+
```
vanitabarrett, line 180: This was an existing section which I've moved to be nested underneath instructions on how to 'Add the JavaScript file to your HTML' because it seems more relevant there, especially as the section on importing using bundlers already contains code examples for initialising individually. Open to any other suggestions.

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -164,36 +164,73 @@ Then import the JavaScript file before the closing `</body>` tag of your HTML pa
 </body>
 ```
 
+#### Select and initialise an individual component
+
+You can select and initialise a specific component by using its `data-module` attribute. For example, initialise the first radio component on a page using `govuk-radios`:
```
EoinShaughnessy, line 169:
```suggestion
You can select and initialise a specific component by using its `data-module` attribute. For example, use `govuk-radios` to initialise the first radio component on a page:
```

EoinShaughnessy (commented):
Left one comment, looks good from a content POV!

source/importing-css-assets-and-javascript/index.html.md.erb
domoscargin: Are we fairly confident that this is pretty much exactly what users will need to add, or should we soften it to something like "...some or all of the following..."?

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -164,36 +164,73 @@ Then import the JavaScript file before the closing `</body>` tag of your HTML pa
...
+    if ($radio) {
+      new Radios($radio).init()
+    }
+  </script>
+```
+
 ### Import JavaScript using a bundler
 
-If you decide to import using a bundler, use `import` to import all of GOV.UK Frontend's components, then run the `initAll` function to initialise them:
+If you decide to import using a bundler, we recommend you use `import` to only import the JavaScript for components you're using in your service. For example:
```
domoscargin, line 183: 🙌

domoscargin:
This looks grand to me. I've added one comment, but it's very minor (and possibly not relevant).

source/importing-css-assets-and-javascript/index.html.md.erb
vanitabarrett: Yeah, good point, I wouldn't say I'm confident on anything Webpack related. I'm pretty sure this bit is needed (both you and Jani needed to add it when testing with Webpack), but there may be additional config not included here. I'll try to soften it a bit.

domoscargin (approved):

PR #180, opened 2022-04-13
Make example heading colons conditional on a description being present

### LINKED ISSUE #117: Remove redundant colons from code example titles in Sass API Reference
## What
Remove redundant colons from code example titles in our [Sass API Reference](https://frontend.design-system.service.gov.uk/sass-api-reference/#sass-api-reference).

## Why
It's inconsistent with our practice elsewhere in the Sass content. Usually, for a code example title, we just display a word, without punctuation. However, there are some instances where a colon displays after the word. For example, there's a colon after the word 'Example' here, but none after the word 'Usage':
image

Laurence, Eoin

## Done when
- [ ] Developer puts colons inside an 'if block,' so that colons only display as part of a description, as here: 
image

querkmachine:
Adds an `if` statement around the colon and description of "example" headings in the Sass API reference, so these only get rendered if a description exists. 

Resolves #117.

### DIFF
--- a/source/sass-api-reference/index.html.md.erb
+++ b/source/sass-api-reference/index.html.md.erb
@@ -56,7 +56,7 @@ $<%= item.context.name %>: <%= item.context.value %>;
 <% if item.example %>
 
 <% item.example.each do |example| %>
-##### Example: <%= example.description %>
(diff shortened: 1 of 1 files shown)


domoscargin (approved):
Neat!

PR #181, opened 2022-04-14
Fix missing links in README

domoscargin:
While setting this repo up on my new machine, I noticed some of the links in the README were out of date.

Most have equivalent documentation in the new tech docs site structure, but I can't find any relevant docs on partials, so I've removed the link for now. It used to live on the [single page documentation](https://github.com/alphagov/tdt-documentation/blob/8549ab8d4b06372acaa69e8e9f9f7f3f9fb9da53/source/create_project/single_page/index.html.md.erb#L36).

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,7 @@ You can add content by editing the `.html.md.erb` files. These files support con
 - HTML
 - Ruby
 
-👉 You can use Markdown and HTML to [generate different content types][example-content] and [Ruby partials to manage content][partials].
-
-👉 Learn more about [producing more complex page structures][multipage] for your website.
+👉 Learn more about [configuring your website][configure].
 
 ## Installing dependencies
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ You can add content by editing the `.html.md.erb` files. These files support con
 - HTML
 - Ruby
 
-👉 You can use Markdown and HTML to [generate different content types][example-content] and [Ruby partials to manage content][partials].
```
vanitabarrett, line 32: Looks like [this file was deleted](https://github.com/alphagov/tdt-documentation/pull/127/files#diff-19535472ebe098550446abc0eeaab92893974cc2a1ce278d4c9db3563bdb2ac0L36) and moved to [a new file](https://github.com/alphagov/tdt-documentation/pull/127/files#diff-8cca47100766286315c6e134732526468608d6197a0015e40c12a5d215aeb008R19) . I'm not sure if it's still relevant for the frontend docs - if not, we could remove this bit.

domoscargin: Yeah, I presume we're referring here to things that look like:

`<% partial foo/bar %>`

Which we don't use in the frontend docs.

I agree though - feels like if we're sending readers to the tech docs documentation anyway, we don't need this bit.

vanitabarrett: Sounds good to me!

domoscargin:
I've replaced two links with a more generic link to the documentation, as discussed.

vanitabarrett (approved):

PR #182, opened 2022-04-14
Add troubleshooting for M1 Macs during bundle install

domoscargin:
During `bundle install` on M1 machines, there appears to be an issue with `eventmachine-1.2.7`, which complains with the error:

```
fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
         ^~~~~~~~~~~~~~~
1 error generated.
```

We've managed to get past this by running `brew install openssl` and following the directions here: https://stackoverflow.com/questions/30818391/gem-eventmachine-fatal-error-openssl-ssl-h-file-not-found/31516586#31516586

This adds these instructions to the troubleshooting section of the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,17 @@ Run `bundle update` to make sure you're using the most recent Ruby gem versions.
 
 Run `bundle exec middleman build --verbose` to get detailed error messages to help with finding the problem.
 
+### OpenSSL issues on newer versions of MacOS
+Because of some quirks with OpenSSL on newer versions of MacOS, you might run into an error with `eventmachine` while running `bundle install`.
+
+To fix this, run:
+
+```
+brew install openssl
+bundle config build.eventmachine --with-cppflags=-I$(brew --prefix openssl)/include
+bundle install
+```
+
 ## Testing
 
 To run the linter and Ruby tests, run `bundle exec rake`.


vanitabarrett:
@querkmachine Might be one for you to review, as you had the same issue earlier this week?

querkmachine:
The instructions themselves look good to me. Some digging I did after-the-fact indicates this is not an M1/Apple Silicon-specific problem*, so the text may need amending to indicate that. 

\* The best explanation I could find is that macOS no longer has OpenSSL bundled with it like past versions, or rather, nowadays OpenSSL only exists as a symlink to LibreSSL—a stripped-down fork of OpenSSL—which isn't working in this instance as eventmachine is looking for a specific file within OpenSSL.

domoscargin:
> Some digging I did after-the-fact indicates this is not an M1/Apple Silicon-specific problem*, so the text may need amending to indicate that.

I'll lower my recently-crafted M1-problems pitchfork and light it less immediately next time!

querkmachine (approved):

lfdebrux:
Interesting, I think this issue should affect all tech docs template instances, as `eventmachine` is a transitive dependency of `govuk-tech-docs`.

Do you think it would be worth raising an issue against https://github.com/alphagov/tech-docs-gem to share the workaround?

PR #186, opened 2022-05-10
Strengthen message around LibSass and Ruby Sass

vanitabarrett:
Make it more explicit that GOV.UK Frontend will drop support for LibSass and Ruby Sass at some point in the future.

### DIFF
--- a/source/installing-with-npm/index.html.md.erb
+++ b/source/installing-with-npm/index.html.md.erb
@@ -21,14 +21,9 @@ weight: 10
 
     If you're using Dart Sass 1.33.0 or greater, you may see deprecation warnings when compiling your Sass. You can [silence deprecation warnings caused by dependencies](/importing-css-assets-and-javascript/#silence-deprecation-warnings-from-dependencies-in-dart-sass) if required.
 
-    We recommend you do not use either LibSass or Ruby Sass, as they are deprecated.
+    Do not use either LibSass or Ruby Sass, which are deprecated, for new projects.
 
-    Currently, GOV.UK Frontend supports:
-
-- LibSass - version 3.3.0 and greater
-- Ruby Sass - version 3.4.0 and greater
-
-    However, we may stop supporting LibSass and Ruby Sass in future.
+    Although GOV.UK Frontend currently supports LibSass (version 3.3.0 and above) and Ruby Sass (version 3.4.0 and above), we will remove support in future. If you're using either of these Sass compilers, you should [migrate to Dart Sass](https://sass-lang.com/blog/libsass-is-deprecated#how-do-i-migrate) as soon as you reasonably can.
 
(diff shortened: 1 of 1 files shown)


EoinShaughnessy (approved):

source/installing-with-npm/index.html.md.erb
```diff
@@ -21,14 +21,9 @@ weight: 10
 
     If you're using Dart Sass 1.33.0 or greater, you may see deprecation warnings when compiling your Sass. You can [silence deprecation warnings caused by dependencies](/importing-css-assets-and-javascript/#silence-deprecation-warnings-from-dependencies-in-dart-sass) if required.
 
-    We recommend you do not use either LibSass or Ruby Sass, as they are deprecated.
+    We strongly recommend you do not use either LibSass or Ruby Sass, as they are deprecated.
```
36degrees, line 24: I think if we make this about 'new projects' (where there won't be an existing build pipeline etc) we can be even harder line on this.

```suggestion
    Do not use either LibSass or Ruby Sass for new projects, as they are deprecated.
```

source/installing-with-npm/index.html.md.erb
```diff
@@ -21,14 +21,9 @@ weight: 10
...
 
-    Currently, GOV.UK Frontend supports:
-
-- LibSass - version 3.3.0 and greater
-- Ruby Sass - version 3.4.0 and greater
-
-    However, we may stop supporting LibSass and Ruby Sass in future.
+    Although GOV.UK Frontend currently supports LibSass (version 3.3.0 and above) and Ruby Sass (version 3.4.0 and above), we will remove support in future.
```
36degrees, line 26:
```suggestion
    Although GOV.UK Frontend currently supports LibSass (version 3.3.0 and above) and Ruby Sass (version 3.4.0 and above), we will remove support in future. If you're using either of these Sass compilers, you should [migrate to Dart Sass](https://sass-lang.com/blog/libsass-is-deprecated#how-do-i-migrate) as soon as you reasonably can.
```

36degrees, line 26: Worth making it explicit what we want service teams to do, and linking to the guidance from the Sass team?

```suggestion
    Although GOV.UK Frontend currently supports LibSass (version 3.3.0 and above) and Ruby Sass (version 3.4.0 and above), we will remove support in future. If you're using either of these Sass compilers, you should [migrate to Dart Sass](https://sass-lang.com/blog/libsass-is-deprecated#how-do-i-migrate) as soon as you reasonably can.
```

36degrees (commented):
Looks good to me. I've added a couple of suggestions, but feel free to ignore them if you don't think they improve things.

source/installing-with-npm/index.html.md.erb
EoinShaughnessy: Really good shout. Suggesting a tweak to be super-clear about what's been deprecated.

```suggestion
    As LibSass and Ruby Sass are deprecated, do not use them for new projects.
```

PR #217, opened 2022-11-01
Update all dependencies using Dependabot

colinrotherham:
This PR enables Dependabot updates for npm, Ruby gems and GitHub Actions

Previously only security-related updates were included

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,13 @@ updates:
     allow:
       - dependency-type: all
 
+  # Update Ruby gems
+  - package-ecosystem: bundler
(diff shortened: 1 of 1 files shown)


colinrotherham:
Questions:

1. Do we want `weekly` not `daily` checks?
2. Do other "worrying updates" need ignoring?
3. Do semver ranges in `package.json` need updating to match?
4. Do we want an ["Add to project" GitHub Action](https://github.com/actions/add-to-project) workflow before we merge?

For 3) this is assuming that Dependabot might respect our configured ranges

36degrees (approved):

PR #220, opened 2022-11-11
Add deployment docs

36degrees:


romaricpascal (approved):

romaricpascal (changes requested):
Actually, found a broken link for the DNS config :( https://github.com/alphagov/govuk-dns-config/blob/master/service.gov.uk.yaml

romaricpascal (approved):

PR #240, opened 2022-11-16
Fix ID generation

romaricpascal:
Temporary fix waiting for a more long term solution to this issue to be implemented in the tech-docs-gem repository: https://github.com/alphagov/tech-docs-gem/issues/310

The broken computation of IDs seems to be due to the the ID generator being reset a bit too loosely. This PR ensures it get reset for the rendering of each page, which makes sure it only considers the IDs on the page itself.

For that it hooks onto [the `preprocess` step of Redcarpet's Markdown rendering](https://github.com/vmg/redcarpet#prepost-process) to reset the ID generator.

The `TechDocsHTMLRenderer` is used in two different places, though (for the [general rendering](https://github.com/alphagov/tech-docs-gem/blob/d3a1545158c5f9c78caafcbbbd7ddd097fdeedcb/lib/govuk_tech_docs.rb#L44) and the [Sass API ref](https://github.com/alphagov/tech-docs-gem/blob/d3a1545158c5f9c78caafcbbbd7ddd097fdeedcb/lib/govuk_tech_docs/api_reference/api_reference_renderer.rb#L142)). This means we cannot extend it and replace with our own renderer.

Instead this PR uses Ruby's ability to add methods to existing classes. To make sure this is done cleanly, [the `preprocess` method is defined in a separate module and then `include`d in the class separately](https://www.justinweiss.com/articles/3-ways-to-monkey-patch-without-making-a-mess/). This:
- gives us a clear hint in stacktraces that its our method that had an issue if that ever happens (rather than pointing to the class where you wouldn't find the method)
- allows us to call `super` if `TechDocsHTMLRenderer` starts implementing a `preprocess` itself (though hopefully it'll be to land the fix, so that'll mean we can get rid of this patch)


colinrotherham:
Nice fix @romaricpascal, is it something we can open as a PR on https://github.com/alphagov/tech-docs-gem?

romaricpascal:
> Nice fix @romaricpascal, is it something we can open as a PR on [alphagov/tech-docs-gem](https://github.com/alphagov/tech-docs-gem)?

I think on the `tech-docs-gem` repo, I'd like to do things a bit more cleanly and stop the ID generator from being a singleton and properly tie it to the markdown rendering. At the moment, it only needs to be a singleton because it is reset and used in a different place.

colinrotherham (approved):
Happy to approve this as a temporary fix (otherwise it'll bite us again)

Some questions, but not blockers:

1. **Shall we wait until it's fixed upstream?**
[alphagov/tech-docs-gem](https://github.com/alphagov/tech-docs-gem)?

2. **What is the patch commit going to solve?**
The commit message doesn't say 😊 

For 1) I'm not sure how long (or complex) an upstream fix is. I'd wait if it's only a quick one, otherwise it's complex and has breaking changes for other teams maybe not

romaricpascal:
Updated the commit message for the patch to quickly describe why and point to the issue in the tech docs repo.

Regarding the first question, this PR is there to have things fixed in our docs waiting for upstream to be fixed, so it wouldn't make sense to wait for it to be fixed before merging 😄

PR #242, opened 2022-11-29
Bump dependencies

domoscargin:
Bumps our npm packages, and our Ruby gems (via `bundle update`).

I've also made our ruby version explicit (there were some failures from Netlify running 2.6.2). I've set it to 2.7.2 to match `tech-doc-gems`, though it's worth mentioning that [2.7 reaches EOL in March](https://endoflife.date/ruby).


romaricpascal:
I'd even be keen to set the ruby version to 3.1 while we're at it, to be ahead rather than at the trail of Ruby version changes. It did work OK locally with 3.1.2 for me, but I think the latest is 3.1.3 (I don't expect much change, though).

domoscargin:
Yeah, think I'll close this one and let Dependabot handle it bit by bit.

PR #252, opened 2022-12-01
Bump npm dependencies

domoscargin:
Attempting to cool Dependabot down a bit on this repo by bumping the npm packages (I'm less familiar with ruby, but might have a look if I find time).

Only major change is `standard`, and our linting checks still pass, so all good.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,10 +10,10 @@
       "hasInstallScript": true,
       "devDependencies": {
         "govuk-frontend": "^4.4.0",
-        "hyperlink": "^5.0.3",
+        "hyperlink": "^5.0.4",
         "sassdoc": "^2.7.4",
-        "standard": "^16.0.4",
-        "tap-spot": "^1.1.1"
+        "standard": "^17.0.0",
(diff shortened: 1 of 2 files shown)


colinrotherham (approved):
Thanks @domoscargin 🎉 

Only bonus suggestion would be to add an ESLint config (now that Standard uses it) so code editors and IDEs can automatically pick up linting in the UI with the right plugins

* https://github.com/alphagov/govuk-frontend/pull/2849/commits/c84641f74496cf0995b55d712ae8153b70ecf504

Not a blocker

domoscargin:
> Thanks @domoscargin 🎉
> 
> Only bonus suggestion would be to add an ESLint config (now that Standard uses it) so code editors and IDEs can automatically pick up linting in the UI with the right plugins
> 
> * [alphagov/govuk-frontend@c84641f](https://github.com/alphagov/govuk-frontend/commit/c84641f74496cf0995b55d712ae8153b70ecf504)
> 
> Not a blocker

Yeah, this is my next move when I get a chance - it would unlock

PR #271, opened 2022-12-16
Load our govuk-frontend package before the tech_docs_gem

romaricpascal:
This makes Sprockets point to whichever version we defined in this repository rather that the one set in its own dependencies and [loaded there](https://github.com/alphagov/tech-docs-gem/blob/d3a1545158c5f9c78caafcbbbd7ddd097fdeedcb/lib/govuk_tech_docs.rb\#L35).

This ensure we're using styles and scripts from the version we set and saves us waiting for the tech_docs_gem to have updated.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -10,6 +10,11 @@ config[:build_dir] = "deploy/public"
 
 GovukTechDocs.configure(self)
 
+# Load our own version of GOV.UK Frontend before the one registered by the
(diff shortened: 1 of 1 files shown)


domoscargin (approved):
Looks sensible to me!

colinrotherham (approved):

PR #276, opened 2022-12-21
Specify Ruby version

domoscargin:
For Netlify builds, some of our dependencies are [complaining about the Ruby version](https://github.com/alphagov/govuk-frontend-docs/pull/267):

`activesupport-7.0.4 requires ruby version >= 2.7.0, which is incompatible with the current version, ruby 2.6.2p47`

Which is fair, since 2.6.2 is EOL!

## Specifying a Ruby version
Netlify allows specification of the ruby version either through setting an environment variable, or by adding a `.ruby-version` file. This PR implements the latter.

I've specified the latest "normal maintenance" version of Ruby: 3.1.3.

## Rubocop offenses
As a result of bumping Ruby, I've had to update `rubocop-govuk` - the version we were using was only compatible with Ruby up to 2.7, which has entered "security maintenance".

This in turn has thrown up a bunch of rubocop offenses. I've run a safe autocorrect to address most of these. For the rest, I've currently disabled the rule which was causing trouble, but would appreciate somebody with Ruby knowledge to take a look and see if we can quickly fix those. I'll leave comments on the relevant lines.


spec/sassdocs_helpers_spec.rb
```diff
@@ -13,10 +13,13 @@ def dothash(hash)
   before(:each) do
     allow(File).to receive(:read).and_return('{ "dependencies": { "govuk-frontend": { "version": "1.0.0" } } }')
     # Include mixin into a test class to allow us to mock File
+    # TODO Move constant definition
+    # rubocop:disable Lint/ConstantDefinitionInBlock
```
domoscargin, line 17: Can this constant definition be edited?
https://msp-greg.github.io/rubocop/RuboCop/Cop/Lint/ConstantDefinitionInBlock.html

Rakefile
```diff
@@ -2,6 +2,7 @@ require "rspec/core/rake_task"
 
 RSpec::Core::RakeTask.new(:spec)
 
+desc "Lint ruby and JavaScript files"
```
colinrotherham, line 5: Probably should title case Ruby like JavaScript

```suggestion
desc "Lint Ruby and JavaScript files"
```

spec/spec_helper.rb
```diff
@@ -44,57 +44,55 @@
...
-  #     --seed 1234
-  config.order = :random
-
-  # Seed global randomization in this process using the `--seed` CLI option.
-  # Setting this allows you to use `--seed` to deterministically reproduce
-  # test failures related to randomization by passing the same `--seed` value
-  # as the one that triggered the failure.
-  Kernel.srand config.seed
-=end
+  # The settings below are suggested to provide a good initial experience
```
colinrotherham, line 47: @domoscargin I feel like we should ignore changes to this file give then message at the top?

```rb
# This file was generated by the `rspec --init` command. Conventionally, all
# specs live under a `spec` directory, which RSpec adds to the `$LOAD_PATH`.
# The generated `.rspec` file contains `--require spec_helper` which will cause
# this file to always be loaded, without a need to explicitly require it in any
# files.
```

colinrotherham: Looks like [`rspec`](https://github.com/rspec/rspec-metagem#readme) has been replaced with [`rspec-rails`](https://github.com/rspec/rspec-rails#readme)

So yeah probably leave this file alone (ignore comment?) and plan for an update one day

># RSpec
>
>Behaviour Driven Development for Ruby
>
>**The rspec metagem repository has been renamed to rspec-metagem, please update
any rspec/rspec Github references to rspec/rspec-metagem, this is in preparation
for a new mono-repo approach to RSpec dev to unify issue tracking and PR management**

colinrotherham (approved):
Approved with some tiny comments

lib/sassdocs_helpers.rb
```diff
@@ -41,6 +41,8 @@ def inline_parameters(parameters)
     # Generate the possible parameters that could be passed into a function or mixin.
     # For example with the `govuk-colour` function: `$colour, $legacy`
     # If the function / mixin has default values output them too, for example `$important: true`
+    # TODO Replace concatenation with interpolation
+    # rubocop:disable Style/StringConcatenation
```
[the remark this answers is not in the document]
domoscargin: Going to leave these as disabled for now.

spec/sassdocs_helpers_spec.rb
domoscargin: Going to leave these as disabled for now.

colinrotherham: I was happy with these TODO comments by the way 👍

spec/spec_helper.rb
domoscargin: Donezos

Issue #308, opened 2023-04-14
Update installation instruction to reflect new JavaScript loading strategy


romaricpascal:
## What

Update the [Get started](https://frontend.design-system.service.gov.uk/get-started/#5-get-the-javascript-working) and [Import CSS, assets and JavaScript](https://frontend.design-system.service.gov.uk/importing-css-assets-and-javascript/#javascript) section to reflect our new way to load JavaScript.

## Why

We're [changing our recommendations for loading our JavaScript](https://github.com/alphagov/govuk-frontend/issues/2639)
and we need to communicate to users how we intend JavaScript to be loaded to ensure they can update their projects accordingly.

## Who needs to know about this

Developers

## Done when

```[tasklist]
- [ ] https://github.com/alphagov/govuk-frontend-docs/issues/203
- [x] Update (draft) Frontend Docs "Get the JavaScript working" to swap `all.js` with ES modules
- [x] Update (draft) Frontend Docs "Import CSS, assets and JavaScript" to swap `all.js` with ES modules
```

colinrotherham:
Forgot this one is still in review:

* https://github.com/alphagov/govuk-frontend-docs/issues/203

romaricpascal:
Moving to "Blocked" as the PRs have been reviewed but are held by other technical PRs (https://github.com/alphagov/govuk-frontend-docs/pull/318, https://github.com/alphagov/govuk-frontend-docs/pull/336 and https://github.com/alphagov/govuk-frontend-docs/pull/335)  they depend on before they can be merged into release/5.0 ultimately.

colinrotherham:
Do we have to block this? I'm happy reviewing documentation knowing it's going into a queue

colinrotherham:
Going to move back into **Needs review**

If you find anything held up let me know, I'll re-order the PRs

colinrotherham:
Closing as we've now [completed those PRs](https://github.com/alphagov/govuk-frontend-docs/issues/308#issuecomment-1634443716) and we have [duplicate CSP hash tasks](https://github.com/alphagov/govuk-frontend/issues/3508)

PR #327, opened 2023-07-18
Preview v5: Update examples using ES modules

colinrotherham:
Aligns all our examples to use `<script type="module">` for https://github.com/alphagov/govuk-frontend-docs/issues/308

Need to consider (maybe) adding a UMD example for those that need `window.GOVUKFrontend` support

### DIFF
--- a/source/configure-components-with-javascript/index.html.md.erb
+++ b/source/configure-components-with-javascript/index.html.md.erb
@@ -73,7 +73,7 @@ new CharacterCount($element, {
       one: "%{count} character to go"
     }
   }
-}).init()
+})
 ```
 
 Read the [JavaScript API Reference](/javascript-api-reference/) to see what configuration each component accepts.
--- a/source/configure-components-with-javascript/index.html.md.erb
+++ b/source/configure-components-with-javascript/index.html.md.erb
@@ -83,7 +83,7 @@ Read the [JavaScript API Reference](/javascript-api-reference/) to see what conf
 You can pass configuration for components when initialising GOV.UK Frontend using the `initAll` function. You can do this by including key-value pairs of camel-cased component names and configuration objects. This is the same method you would use to pass them when creating an instance of the component. For example:
 
 ```javascript
-window.GOVUKFrontend.initAll({
+initAll({
   characterCount: {
     maxlength: 500, // Non namespaced
     i18n: { // i18n namespace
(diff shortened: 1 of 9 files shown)


source/configure-components-with-javascript/index.html.md.erb
```diff
@@ -83,7 +83,7 @@ Read the [JavaScript API Reference](/javascript-api-reference/) to see what conf
 You can pass configuration for components when initialising GOV.UK Frontend using the `initAll` function. You can do this by including key-value pairs of camel-cased component names and configuration objects. This is the same method you would use to pass them when creating an instance of the component. For example:
 
 ```javascript
-window.GOVUKFrontend.initAll({
+initAll({
```
domoscargin, line 86: What contexts will `initAll` be available in? Does this need an import statement to make it clear?

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -185,17 +186,15 @@ If you decide to import using a bundler, we recommend you use `import` to only i
...
 import { SkipLink, Radios } from 'govuk-frontend'
 
-var $skipLink = document.querySelector('[data-module="govuk-skip-link"]')
+const $skipLink = document.querySelector('[data-module="govuk-skip-link"]')
 if ($skipLink) {
   new SkipLink($skipLink).init()
 }
 
-var $radios = document.querySelectorAll('[data-module="govuk-radios]')
-if ($radios) {
```
domoscargin, line 194: Huh, nice catch here on the return value of `querySelectorAll`!

domoscargin (commented):
Looks good.

Is there somewhere we can just add a generic and brief "if you need to use `window.GOVUKFrontend`" or "if you're using UMD" section with a small explanation?

I'm not up to date with what proportion of users we think are going to need this functionality - it'd be good to head support requests off with some documentation if we think it'll be significant.

source/configure-components-with-javascript/index.html.md.erb
colinrotherham: Thanks for spotting this @domoscargin, makes me feel better

I had the same worries but noticed we (mostly) don't add imports in other examples

But otherwise, we'd have to weigh up three approaches now:

1. UMD via `const { initAll } = require('govuk-frontend')`
1. ES modules (bundler) via `import { initAll } from 'govuk-frontend'`
1. ES modules (browser) via `import { initAll } from '/javascripts/govuk-frontend-5.0.0.min.js'`

One to improve in future?

colinrotherham:
> Is there somewhere we can just add a generic and brief "if you need to use `window.GOVUKFrontend`" or "if you're using UMD" section with a small explanation?

Great minds! I'd added this to @claireashworth's content document and we should do some more thinking, perhaps similar to the [**Support Internet Explorer 8**](https://frontend.design-system.service.gov.uk/supporting-ie8/) page? For those that need the window globals (or UMD) still

I'd like to raise a separate PR for it

source/configure-components-with-javascript/index.html.md.erb
domoscargin: Yeah, that sounds sensible!

domoscargin (approved):
Hot to trot.

PR #328, opened 2023-07-18
Add an ERB comment pointing to the govuk-frontend test showing the hash

romaricpascal:
This should make it easier for making or reviewing future updates of the frontend docs 😊

### DIFF
--- a/source/importing-css-assets-and-javascript/index.html.md.erb
+++ b/source/importing-css-assets-and-javascript/index.html.md.erb
@@ -245,6 +245,10 @@ Make sure you [understand the security implications of using either option](http
 #### Use a hash to unblock inline JavaScript
 
 You can unblock inline JavaScript by including the following hash in your CSP:
+<%#
+  This hash should match the one in `govuk-frontend` tests:
+  https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/template.test.js
+%>
 
 ```
 sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU=


source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -245,6 +245,10 @@ Make sure you [understand the security implications of using either option](http
 #### Use a hash to unblock inline JavaScript
 
 You can unblock inline JavaScript by including the following hash in your CSP:
+<%#
+  This hash should match the one in `govuk-frontend` tests:
+  https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/template.test.js#L165-L175
```
36degrees, line 250: I'm not sure it's worth including the line numbers, simply because as the file changes it could end up pointing to the wrong place in the file (which may confuse people)

Equally I don't think we want to use a permalink here as we want to make sure people look at the latest version of the file.

```suggestion
  https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/template.test.js
```

36degrees (approved):
Thanks for doing this! Minor non-blocking suggestion.

source/importing-css-assets-and-javascript/index.html.md.erb
romaricpascal: Fair point on the line numbers, I'll remove them. I don't understand what you mean by "using a permalink" and "we want to make sure people look at the latest version of the file", though, sorry. Can you elaborate, please?

36degrees: Sure thing – if you use the little `…` on the line numbers in the file you get the option to copy a permalink[^1]

Screenshot 2023-07-20 at 10 02 38

That will give you a link that includes the sha for the specific commit you were looking at (like https://github.com/alphagov/govuk-frontend/blob/cf1eaf0423d304fdb47f34ffa152659b6ff6275d/packages/govuk-frontend/src/govuk/template.test.js#L165-L175) rather than referencing a branch, which will always reflect the current state of that branch.

That would mean that we could use line numbers, because the link will always reference `cf1eaf0` which won't change.

But users may not realise they're looking at the code at a fixed point in the past, and assume the hash they're seeing in the tests is the 'current' hash.

So I think pointing to the `main` branch is the right thing to do in this instance.

[^1]: **Pro-tip:** You can also just press <kbd>y</kbd> anywhere to change the URL of the page to a permalink

romaricpascal: Oh I see, yeah, I did a mix between permalink for the lines and a link to the file on the branch, which is not helpful. I'll just point to main 😊Thanks for the explanation and the tip.

PR #332, opened 2023-07-21
Preview v5: Content changes

claireashworth:
Removes references to legacy frameworks, supporting IE8, minimum Node.js version etc


source/supporting-ie8/index.html.md.erb
```diff
@@ -1,131 +0,0 @@
----
-title: Support Internet Explorer 8
```
colinrotherham, line 2: 🙌

colinrotherham (approved):
Happy to approve. Probably hold off merging until we've got a v5 release candidate of `govuk-frontend` to install into the docs website 🎉

claireashworth:
Thanks. 😁 My rough plan is to make all the content changes here so they're ready to go when everything else is set up as well.

colinrotherham:
Hi @claireashworth

I've dropped the commits that deleted these pages:

```
source/compatibility-mode/index.html.md.erb
source/migrating-from-legacy-products/index.html.md.erb
source/supporting-ie8/index.html.md.erb
```

Let's find out if we need to follow the [**Archiving deleted URLs**](https://github.com/alphagov/govuk-design-system/blob/main/docs/contributing/archiving-deleted-urls.md) guidance

If so, we have slightly different options on this repo:

```md
---
title: {Title of the page you are archiving}
hide_from_sitemap: true
hide_in_navigation: true
prevent_indexing: true
---
```

36degrees:
> I've dropped the commits that deleted these pages:
> 
> ```
> source/compatibility-mode/index.html.md.erb
> source/migrating-from-legacy-products/index.html.md.erb
> source/supporting-ie8/index.html.md.erb
> ```
> 
> Let's find out if we need to follow the [**Archiving deleted URLs**](https://github.com/alphagov/govuk-design-system/blob/main/docs/contributing/archiving-deleted-urls.md) guidance

Agree that we shouldn't just delete the pages. I think we might want to either:

- redirect the pages to the v4 version (which might not be obvious enough to users)
- replace them with a non-indexed, non-linked page that explains that to use compatibility mode / support IE8 they need to use v4, then link to the v4 version of the guidance

We may need to look at #346 first before we can answer this properly?

colinrotherham:
@36degrees To unblock this PR, shall we keep @claireashworth's non-indexed page content for now?

I can help rename them to their original names so we meet this suggestion:

> * replace them with a non-indexed, non-linked page that explains that to use compatibility mode / support IE8 they need to use v4, then link to the v4 version of the guidance

Can look at links/redirects or other improvements after https://github.com/alphagov/govuk-frontend-docs/issues/346?

romaricpascal:
Ahead of the pre-release, I think we can quickly go with:

> replace them with a non-indexed, non-linked page that explains that to use compatibility mode / support IE8 they need to use v4, then link to the v4 version of the guidance

I'm not sure we even need to update any content on these pages for now, just prevent them from being linked and indexed.

Once we've sorted the technicalities for hosting v4 in #346, we can look at proper redirects.

colinrotherham:
Forgot to say @claireashworth I did this yesterday

Added a minor tweak for Node.JS → Node.js but we do that often elsewhere too (use the wrong one)

colinrotherham (approved):

PR #335, opened 2023-07-28
Preview v5: GitHub Actions fixes

colinrotherham:
This PR includes a couple of GitHub Actions fixes:

1. The npm script `check-links` can fail silently
2. The checks are duplicates across `pull` and `pull_request` events

## Broken image
Matching the [GOV.UK Design System `check-links` arguments](https://github.com/alphagov/govuk-design-system/commit/85371227743245c38c6a25c89eb0e3b2157f8174) found a `--skip` argument typo

We have a broken image that should skipped with ~`property="og:url"`~ `property="og:image"`

* https://github.com/alphagov/tech-docs-gem/issues/194

PR #343, opened 2023-08-16
Fix link checker reporting

36degrees:
**This PR cherry-picks 3e3adb6e2f28aa14e7706fecce81b34b11f5d967 from #335 into the main branch to unblock merging PRs to main because of link checks failing ([Example](https://github.com/alphagov/govuk-frontend-docs/actions/runs/5877869593/job/15943224377?pr=338))**

---

We often see sucessful log output from `tap-spot` followed by a non-zero error code so we’ve swapped it for `tap-mocha-reporter` like we did on GOV.UK Design System

I’ve also added the `--canonicalroot` flag to check more internal assets that would have previously been skipped as external

Note: We still have a broken share image

* https://github.com/alphagov/tech-docs-gem/issues/194

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "hyperlink": "^5.0.4",
         "sassdoc": "^2.7.4",
         "standard": "^17.1.0",
-        "tap-spot": "^1.1.2"
+        "tap-mocha-reporter": "^5.0.3"
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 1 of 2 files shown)


36degrees:
@colinrotherham any objections to this being cherry-picked across to main? There were no conflicts so hoping the v5 branch will still merge cleanly.

colinrotherham (approved):

PR #349, opened 2023-09-27
Correct node_modules folder name in index.html.md.erb

kouphax:
In point 2 of section "5. Get the JavaScript working" there was an incorrect reference to the node_modules folder which was typed as as node-modules (hyphen instead of underscore).

This change fixes that small issue.


colinrotherham (approved):

domoscargin (approved):

colinrotherham:
Thanks @kouphax, well spotted

Issue #341, opened 2023-08-08
Update JavaScript content in Frontend for v5 changes


claireashworth:
## What
Frontend currently mainly has JavaScript information in 3 pages:

- Import CSS, assets and JavaScript
- Configure components with JavaScript
- JavaScript API Reference

Discuss and decide what content changes are needed and in which pages they should be. Also, discuss the potential for a new page that either brings together the JavaScript info spread across the pages or is specifically about 'Initialising JavaScript'.

Then:

- Draft content changes into agreed locations
- Update cross-links as needed
- Review and publish as part of v5

## Why
This is part of the changes to the v5 release

## Who needs to know about this
Tech writer, developers

## Done when

- [ ] Arrange team meeting to discuss JavaScript content updates
- [ ] Draft agreed content
- [ ] Review and publish with v5

romaricpascal:
Overlaps with #373 and #380.

PR #352, opened 2023-10-02
Add explanation that individual initialisation will throw errors

romaricpascal:
As we decided that to start with, [we’d limit our public API to “we throw errors”](https://github.com/alphagov/govuk-frontend/issues/4074#issuecomment-1731127591), we don’t have any specific errors to document as part of our [JavaScript API Reference](https://frontend.design-system.service.gov.uk/javascript-api-reference/). 

I think we don't need more than [an update to “Select and initialise individual components” (Netlify preview)](https://document-errors--govuk-frontend-docs-preview.netlify.app/importing-css-assets-and-javascript/#select-and-initialise-an-individual-component), which this PR adds.

Only other thing I could think of may be to discuss “Initialising components” to the JavaScript API Reference to explain that initAll doesn’t throw but logs and individual initialisation throws, but there’s a bit of overlap with the [JavaScript section of Import CSS, assets and JavaScript](https://document-errors--govuk-frontend-docs-preview.netlify.app/importing-css-assets-and-javascript/#javascript) that lists all the options for initialising already. Seems like this would be a wider piece of work around structuring how we organise the JavaScript API docs (especially to keep the component names in the menu).

Part of the work for #341

### DIFF
--- a/source/importing-css-assets-and-javascript/index.html.md.erb
+++ b/source/importing-css-assets-and-javascript/index.html.md.erb
@@ -162,6 +162,16 @@ Then import the JavaScript file before the closing `</body>` tag of your HTML pa
 </body>
 ```
 
+Errors from components will be logged in the browser's console.
+
+For example, when:
+
+* GOV.UK Frontend is not supported in the current browser
+* Component templates have missing changes from our release notes
+* Component JavaScript configuration does not match our documentation
+
+You should check your application works without errors or some components will not work correctly.
+
 #### Select and initialise an individual component
 
 You can select and initialise a specific component by using its `data-module` attribute. For example, use `govuk-radios` to initialise the first radio component on a page:
@@ -175,6 +185,8 @@ You can select and initialise a specific component by using its `data-module` at
     }
   </script>
 ```
+
+When initialised individually, errors are thrown rather than logged. You must check your application works without errors or some components will not work correctly.
 
 ### Import JavaScript using a bundler
 


source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -176,6 +176,13 @@ You can select and initialise a specific component by using its `data-module` at
   </script>
 ```
 
+When initialised individually, components will throw errors if GOV.UK Frontend is not supported, some key markup is missing or their configuration is invalid.
```
colinrotherham, line 179: Would you mind if we moved this whole section up where `initAll()` is the default?

This current section is optional and might be missed, but we want people to notice the good stuff 🙌

For most of our users on `initAll()` we could list some examples logged to the console:

>Errors from components will be logged in the browser's console.
>
>For example, when:
>
>* GOV.UK Frontend is not supported in the current browser
>* Component templates have missing changes from our release notes
>* Component JavaScript configuration does not match our documentation

Which means we can now keep this section simple:

```suggestion
When initialised individually, errors are thrown rather than logged and must be handled with a [`try/catch` block](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch).
```

colinrotherham: Do you mind if we don't give developer users the option to let their service crash?

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -162,6 +162,14 @@ Then import the JavaScript file before the closing `</body>` tag of your HTML pa
...
 ```
 
+Errors from components will be logged in the browser's console.
+
+For example, when:
+
+* GOV.UK Frontend is not supported in the current browser
+* Component templates have missing changes from our release notes
+* Component JavaScript configuration does not match our documentation
+
```
colinrotherham, line 172: I like the new addition regarding execution stopping, but fixing errors applies to this section too?

```suggestion
* Component JavaScript configuration does not match our documentation

You must check your application works runs without errors or some components will not work correctly.
```

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -176,6 +184,12 @@ You can select and initialise a specific component by using its `data-module` at
   </script>
 ```
 
+When initialised individually, errors are thrown rather than logged.
+This will stop the execution of the script and you'll need to either:
+
+* fix what causes the error
+* use a [`try/catch` block](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch) to change the behaviour of your code in reaction to the error
```
colinrotherham, line 191: With the section above saying "You must check your application works without errors" we don't need to repeat ourselves and talk about fixing errors here too

```suggestion
When initialised individually, errors are thrown rather than logged and must be handled with a [`try/catch` block](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch).
```

source/importing-css-assets-and-javascript/index.html.md.erb
colinrotherham: I've used "should" here and "must" below

source/importing-css-assets-and-javascript/index.html.md.erb
romaricpascal: Thinking we benefit from a little duplication of the reminder to fix errors here for people getting sent a link to that section specifically (and who would not have seen the previous mention).

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -162,6 +162,16 @@ Then import the JavaScript file before the closing `</body>` tag of your HTML pa
...
 
+Errors from components will be logged in the browser's console.
+
+For example, when:
+
+* GOV.UK Frontend is not supported in the current browser
+* Component templates have missing changes from our release notes
+* Component JavaScript configuration does not match our documentation
+
+You should check your application works without errors or some components will not work correctly.
```
colinrotherham, line 173: Thanks for adding this in 🎉

For context, it's taken from here (scroll down):
[**get-started**/#5-get-the-javascript-working](https://deploy-preview-333--govuk-frontend-docs-preview.netlify.app/get-started/#5-get-the-javascript-working)

>In your live application:
>
>- you must use `initAll` to initialise all components that use GOV.UK Frontend's JavaScript, or some components will not work correctly for disabled users who use assistive technologies

source/importing-css-assets-and-javascript/index.html.md.erb
colinrotherham: Yeah that sounds good

The current wording says they can pick either fixing errors or adding `try…catch` but not both

How about we duplicate the one-liner above but change _should_ to _must_?

```suggestion
You must check your application works without errors or some components will not work correctly.
```

<br>

We can always mention the `try…catch` as an optional safety feature rather than something that’s mandatory. For example, their test coverage might not hit every scenario where an error is thrown.

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -176,6 +186,9 @@ You can select and initialise a specific component by using its `data-module` at
   </script>
 ```
 
+When initialised individually, errors are thrown rather than logged.
+You should check your application works without errors or some components will not work correctly.
```
colinrotherham, line 190: Cheers that’s loads simpler

Want to upgrade that “should” to a “must” in this case?

```suggestion
When initialised individually, errors are thrown rather than logged. You must check your application works without errors or some components will not work correctly.
```

source/importing-css-assets-and-javascript/index.html.md.erb
colinrotherham: Brill, thanks for keeping it simple

source/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -176,6 +186,9 @@ You can select and initialise a specific component by using its `data-module` at
   </script>
 ```
 
+When initialised individually, errors are thrown rather than logged.
+You must check your application works without errors or some components will not work correctly.
```
colinrotherham, line 190: Cheers for updating with ~"should"~ "must"

Did you want these lines to flow into each other?

Markdown will need another new line otherwise this one gets removed

```suggestion
When initialised individually, errors are thrown rather than logged.

You must check your application works without errors or some components will not work correctly.
```

```suggestion
When initialised individually, errors are thrown rather than logged. You must check your application works without errors or some components will not work correctly.
```

colinrotherham (approved):
Looks good. Minor [formatting comment](https://github.com/alphagov/govuk-frontend-docs/pull/352#pullrequestreview-1652881009) but not a blocker

romaricpascal:
@colinrotherham Cheers! Updated.
@claireashworth Does it all look OK from a content perspective as well? 😊

claireashworth:
Looks good to me

PR #353, opened 2023-10-03
Use Node.js v18 + `"lockfileVersion": 3` etc

colinrotherham:
This PR splits out a few maintenance changes from [`release/5.0`](https://github.com/alphagov/govuk-frontend-docs/tree/release/5.0)

1. Use Node.js LTS ‘Hydrogen’ (v18) release version
2. Convert **package-lock.json** to `"lockfileVersion": 3`
3. Prevent GitHub Actions [**Test** workflow](https://github.com/alphagov/govuk-frontend-docs/actions/workflows/test.yaml) duplicate runs (on push)
4. Allow manual GitHub Actions [**Test** workflow](https://github.com/alphagov/govuk-frontend-docs/actions/workflows/test.yaml) runs

After merging Dependabot updates, item 2) causes conflicts with [`release/5.0`](https://github.com/alphagov/govuk-frontend-docs/tree/release/5.0) so best to merge before Monday


.github/workflows/test.yaml
```diff
@@ -2,9 +2,6 @@ name: Test
 
 on:
   pull_request:
-  push:
-    branches-ignore:
-      - main  # pushes to main are handled by deploy.yaml
```
romaricpascal, line 7: **question** The comment here makes it sound like it prevented a double run on push already, from both `test`, and `deploy` (which runs `test` as well). What changed to make it unnecessary?

colinrotherham: Nothing changed, it just got a bit annoying

Every time you push to a PR branch it runs all the checks for both `pull_request` and `push` 😣

The comment shows we only de-duplicated `main` branch in the past

colinrotherham: See also:

* https://github.com/alphagov/govuk-frontend/commit/4690934fac627eedeb79208569719fb99a4953a8
* https://github.com/alphagov/govuk-design-system/commit/dc6b2d64db5e6584d1e8a9c09391650da2b4ed31

romaricpascal (approved):

PR #360, opened 2023-10-27
Prevent confusion from preview site linking to release site

colinrotherham:
When browsing the [v5 preview site](https://release-5-0--govuk-frontend-docs-preview.netlify.app), some links sneakily direct users to the [v4 release site](https://frontend.design-system.service.gov.uk)


romaricpascal (approved):
Good spot! Thanks for fixing those 😊

PR #363, opened 2023-11-01
Allow Node.js LTS ‘Iron’ (v20) release version

colinrotherham:
This PR allows Node.js LTS ‘Iron’ (v20) to install dependencies for **GOV.UK Frontend Docs**

>GOV.UK Frontend requires Node.js version 12.17.0 or later to support ECMAScript modules. Where possible, we recommend you install the latest Long Term Support (LTS) version.

Changes included:

1. **Updating [.nvmrc](https://github.com/alphagov/govuk-frontend/blob/main/.nvmrc)** for Node.js 20
So tools like [nvm](https://github.com/nvm-sh/nvm#readme) and [asdf](https://github.com/asdf-vm/asdf#readme) allow the `lts/iron` LTS release

3. **Add the [package.json `engines.node` range](https://github.com/alphagov/govuk-frontend/blob/main/package.json#L6)**
Ensures `npm install` and other scripts check the local Node.js version is allowed

3. **Add the [package.json `engines.npm` range](https://github.com/alphagov/govuk-frontend/blob/main/package.json#L7)**
Ensure support for `npm@10` rather than `npm@9`

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -23,6 +23,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           cache: npm
+          check-latest: true
           node-version-file: .nvmrc
 
       - uses: ruby/setup-ruby@v1
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/hydrogen
+lts/iron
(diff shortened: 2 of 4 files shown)


colinrotherham:
Looking good

GOV.UK Frontend still gets installed in the same place

On this repo we only use Node.js for:

1. Building SassDoc via `npm run build:sassdoc`
2. Linting JavaScript via `npm run lint`
3. Checking links via `npm run check-links`

domoscargin (approved):
👍

PR #384, opened 2023-11-21
Host v4 in its own folder

### LINKED ISSUE #346: Spike technical approach for hosting v4 in a subfolder
## What

Explore the impacts of hosting v4 in a subfolder, especially:
- links (eg. [moving to relative links](https://github.com/alphagov/govuk-prototype-kit-docs/pull/84), it might be worth thinking at some linting there as well)
- making the frontend-docs menu work with two sets of documentation (we may want to look into what [publishing docs](https://docs.publishing.service.gov.uk/manual/get-started.html) are doing)
- having a way to navigate between v4 and v5 docs (similarly, looking at publishing docs may be a good start, but we may want to anticipate how to look at more than 2 versions for future upates when we expand our public API, for ex).

## Why

We need to validate the feasibility of our plan for hosting older version of frontend-docs in sub folders

Developers

## Done when

- [ ] We've listed the constraints hosting v4 in a subfolder raises
- [ ] We have a plan to work with each of these constraints

romaricpascal:
Cherry picks the commits from the spike exploring how to host docs for v4 in their own folder, next to those of v5.

The commits paint a good pictures of the changes, but as a summary:

- This moves the v4 content in its own folder with v5 content being at the root of the site.
- Links in the content are now relative, to ensure pages for a given version link to other pages of the same version. This applies to both v4 and v5, so it'll be easier to move v5 docs should we need to move it to its own folder in a future version
- Two separate Sass API reference get generated so each version gets the right content for each.
- Pages for the v4 version get their own layout so they can have a separate table of content.
- The header also contains a new menu allowing to navigate between the versions.

Closes #346.


36degrees:
I've spent a bit of time trying to get this passing on CI. I've forced push some changes that amend some of the existing commits. They're mainly fixing rubocop violations – you can [see the changes here](https://github.com/alphagov/govuk-frontend-docs/compare/40000c4d152478cd858415323c588ed93649f2cf%E2%80%A64162f37b93d5bbbd0df1f7bb356edf6eafaff3ba).

I'm not entirely sure about [this change](https://github.com/alphagov/govuk-frontend-docs/commit/fc8a051d17a1f6a88a76d16545211b21260eca70#diff-65cb06ae362f905e275f229d66769f927cc905d63972d29f83df4c6fa67faaa0R33), which changes `rescue` to `rescue StandardError` to avoid [Rubocop's 'Avoid rescuing without specifying an error class' error](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/RescueStandardError). I don't know if there's a more specific error class we could be rescuing because I'm unsure what it is we're guarding against.

I've also pushed some additional commits to fix a couple of broken links.

The only other thing I've noticed is that the highlighting of the current page / section in the side navigation is broken in the v4 docs, I think related to [some of the logic in the tech docs JavaScript that checks URLs match](https://github.com/alphagov/tech-docs-gem/blob/207bcb8593197f3aad8983018ca0a91db7783410/lib/assets/javascripts/_modules/in-page-navigation.js#L71-L85), but I haven't quite got my head around it yet.

romaricpascal:
Cheers for looking into it 😊  That `rescue` should actually be an `ensure` (aim is to make the equivalent of a `try {} finally {}` to ensure things are reset whatever happens), so that'll avoid the rubocop complaints.

stevenjmesser:
image

I’m loving the version selector so that users can move between documentation for different versions of GOV.‌UK Frontend. So good!

36degrees (approved):

Issue #364, opened 2023-11-03
Configure Netlify to use GOV.UK branded 404 pages


36degrees:
## What

In anticipation of moving the Frontend Docs to Netlify, configure Netlify to use our [custom error page](https://github.com/alphagov/govuk-frontend-docs/blob/d2dfde4e76638e55fd18878f38268cdb738bb6b0/source/errors/404.html.md.erb) for 404s.

For example, visiting https://govuk-frontend-docs-preview.netlify.app/foo should display the same GOV.UK branded 404 page as seen at https://frontend.design-system.service.gov.uk/foo.

Leave the existing nginx config for error pages on PaaS, as it's still needed until we've finished the migration.

## Further reading

- [Existing nginx config for error pages](https://github.com/alphagov/govuk-frontend-docs/blob/d2dfde4e76638e55fd18878f38268cdb738bb6b0/deploy/nginx.conf#L128)
- [Netlify's docs on configuring custom 404 pages](https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling)

## Why

So that anyone encountering a broken link or attempting to access a non-existent page sees a GOV.UK branded page that reassures them they are on the right site, and directs them somewhere useful.

## Who needs to know about this

Developers

## Done when

- [ ] https://govuk-frontend-docs-preview.netlify.app/foo displays a GOV.UK branded 404 page

PR #385, opened 2023-11-23
Set up Netlify custom `404.html` page

colinrotherham:
Moves Nginx `/errors/404.html` to `/404.html` which also closes https://github.com/alphagov/govuk-frontend-docs/issues/364

Give it a try: https://deploy-preview-385--govuk-frontend-docs-preview.netlify.app/foo


romaricpascal (approved):
One step closer to being ready to move 🙌🏻  Thanks 😁

PR #402, opened 2023-12-08
Update JavaScript API Reference for v5

colinrotherham:
This PR simplifies the "JavaScript API Reference" introduction by:

1. Clearly linking to the new "Configure components" page
2. Matching config wording to "Configure components" headings
3. Adding anchor links to each component by name
4. Removing the `Button` example

Regarding 4) we have better `CharacterCount` examples on "Configure components" already

## Configuration objects or options

I've also clarified a few mix-ups between "configuration" and "options" such as:

```patch
- Read the JavaScript API Reference for the configuration accepted by each component.
+ Read the JavaScript API Reference for the list of options in each component configuration.
```

```patch
- Read the JavaScript API Reference for the configuration accepted by `initAll`.
+ Read the JavaScript API Reference for the list of components in the `initAll` configuration.
```

### DIFF
--- a/source/javascript-api-reference/index.html.md.erb
+++ b/source/javascript-api-reference/index.html.md.erb
@@ -5,15 +5,10 @@ weight: 80
 
 # JavaScript API Reference
 
-Some of our components can receive configuration options when you create an
-instance in JavaScript. This page lists the available options for these
-components. You can find further information on how to [configure these options
-in our guidance](../configure-components/).
+Some of our components can be [passed JavaScript configuration](../configure-components/#passing-javascript-configuration) objects as:
 
-You can pass these options in an object either as:
-
-- a second argument of the component's constructor
-- the value for the key of the component name in an object passed to `initAll`
+- the second argument when configuring individual component instances
+- the first argument when configuring all components using the `initAll` function
 
 ## Accordion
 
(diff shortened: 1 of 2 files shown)


source/javascript-api-reference/index.html.md.erb
```diff
@@ -5,32 +5,19 @@ weight: 80
 
 # JavaScript API Reference
 
-Some of our components can receive configuration options when you create an
-instance in JavaScript. This page lists the available options for these
-components. You can find further information on how to [configure these options
-in our guidance](../configure-components/).
+Some of our components can be [passed JavaScript configuration](../configure-components/#passing-javascript-configuration) objects as:
```
romaricpascal, line 8: I think the link points a bit too high in the heading of the other page. Once you're there you have to figure out which sub-section applies. How about pointing to the specific sections with:

```suggestion
Some of our components can be passed JavaScript configuration objects. You can do it both:
- [when configuring individual component instances](../configure-components/#with-configuration-object)
- [when configuring all components using the `initAll` function](../configure-components/#configure-all-components-using-the-initall-function)
```

That would also save us talking about the precision of which argument does what, where "first argument" doesn't quite represent where the configuration for a given component goes (it is the first argument, but under a key for the element name).

romaricpascal (approved):
Thanks for updating! ⛵

PR #419, opened 2024-01-12
Configure Netlify Cache-Control headers

### LINKED ISSUE #383: Add caching headers for fingerprinted assets
## What

Have Netlify respond with the following `Cache-Control` headers for the assets that are fingeprinted (either by the tech_docs_gem, if any, or already fingerprinted ones, like the fonts):

```txt
public,max-age=2147483648,immutable
```

## Why

Beyond being good caching practice for ourselves, we want our sites to be a good example of how to set up headers for assets caching, in case people copy it. 

Fingerprinted assets can be cached forever by the browser and our headers should reflect that.

## Who needs to work on this

Developers

## Who needs to review this

Developers

## Done when

- [ ] Netlify serves fingerprinted assets with the right headers.

36degrees:
This mirrors the changes made in https://github.com/alphagov/govuk-design-system/pull/3417, allowing files with ‘fingerprint’ hashes in their filenames to be treated as immutable and cached indefinitely.

Unfortunately in this repo at the minute that only includes the font assets included from GOV.UK Frontend.

Closes #383

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -18,3 +18,15 @@
   from = "/configure-components-with-javascript/"
   to = "/configure-components/"
   status = 301
+
+# Asset filenames with fingerprint hashes (which at the minute is only the font
+# from GOV.UK Frontend) are given an 'infinite' max-age and treated as
+# immutable.
+#
+# Historically, an 'infinite' max-age is the 32-bit maximum 2,147,483,648.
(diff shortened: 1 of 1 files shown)


romaricpascal (approved):
Looks good! ⛵

PR #443, opened 2024-05-10
Update 'Import JavaScript' page to use new `createAll` function

romaricpascal:
> [!WARNING]
> This PR should only be merged after the next release of GOV.UK Frontend

Updates the Frontend docs to reflect the [addition of a new `createAll` function](https://github.com/alphagov/govuk-frontend/pull/4975) to initialise all instances of a specific component on the page.

Also moves the 'Initialise only part of a page' to the 'Import JavaScript' page, rather than 'Configure components' to regroup both how to import and initialise components.

[Each commit](https://github.com/alphagov/govuk-frontend-docs/pull/443/commits) details the changes of a specific page.

Closes alphagov/govuk-frontend#4945

### DIFF
--- a/source/configure-components/index.html.md.erb
+++ b/source/configure-components/index.html.md.erb
@@ -13,7 +13,7 @@ You can configure a component by:
 - [passing JavaScript configuration](#passing-javascript-configuration)
 - [adding HTML data attributes](#adding-html-data-attributes)
 
-Configuration passed via data attributes in the HTML or via Nunjucks macro options will take precedence over any JavaScript configuration.
+Configuration passed through data attributes in the HTML or Nunjucks macro options will take precedence over any JavaScript configuration.
 
 To learn more about how configuration is passed from Nunjucks macros to HTML data attributes, see advanced examples in [the localisation options](../localise-govuk-frontend/).
 
@@ -72,22 +72,6 @@ initAll({
   }
 })
 ```
-
-#### Initialise only part of a page
-
-If you update a page with new markup, for example a modal dialogue box, you can run `initAll` with a `scope` parameter to initialise the components on part of a page:
-
-```javascript
-import { initAll } from 'govuk-frontend'
-
-const $element = document.querySelector('.app-modal')
-
-if ($element) {
-  initAll({ scope: $element })
-}
-```
-
-Read the [JavaScript API Reference](../javascript-api-reference/) for the list of components in the `initAll` configuration.
 
 ### JavaScript errors in the browser console
 
--- a/source/configure-components/index.html.md.erb
+++ b/source/configure-components/index.html.md.erb
@@ -13,6 +13,10 @@ You can configure a component by:
 - [passing JavaScript configuration](#passing-javascript-configuration)
 - [adding HTML data attributes](#adding-html-data-attributes)
 
+Configuration passed via data attributes in the HTML or via Nunjucks macro options will take precedence over any JavaScript configuration.
+
+To learn more about how configuration is passed from Nunjucks macros to HTML data attributes, see advanced examples in [the localisation options](../localise-govuk-frontend/).
+
 ## Setting Nunjucks macro options
 
 If you're using the Nunjucks macros, you can read about [configuring a component](../use-nunjucks/#configuring-a-component) or find all of the configuration options published on the [GOV.UK Design System website](https://design-system.service.gov.uk/).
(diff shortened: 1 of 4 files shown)


source/configure-components/index.html.md.erb
```diff
@@ -13,6 +13,10 @@ You can configure a component by:
 - [passing JavaScript configuration](#passing-javascript-configuration)
 - [adding HTML data attributes](#adding-html-data-attributes)
 
+Although [JavaScript configuration objects](#passing-javascript-configuration) are optional, configuration can still be provided or overridden by [HTML data attributes](#adding-html-data-attributes).
```
36degrees, line 16: The 'although' here doesn't really make sense any more now that we've moved this. I also think we should talk about Nunjucks options, even if they are using data attributes under the hood. I also think it's expected / implied that they work even if you don't pass JS config, so not sure if we need that?

What about something like:

```suggestion
Configuration passed via data attributes in the HTML or via Nunjucks macro options will take precedence over any JavaScript configuration.
```

source/configure-components/index.html.md.erb
```diff
@@ -23,44 +27,18 @@ Read the [JavaScript API Reference](../javascript-api-reference/) for the list o
...
-
-1. The HTML element that represents the component.
-2. An optional configuration object.
-
-Although JavaScript configuration objects are optional, configuration can still be provided or overridden by [HTML data attributes](#adding-html-data-attributes).
-
-To learn more about how configuration is passed from Nunjucks macros to HTML data attributes, see advanced examples in [the localisation options](../localise-govuk-frontend/).
-
-#### Without configuration object
+### Configure instances of a specific component
```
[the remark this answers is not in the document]
36degrees: Do we want to match the format used for `initAll`?

```suggestion
### Configure instances of a specific component using the `createAll` function
```

(And update the `initAll` heading to wrap `initAll` in code blocks?)

source/configure-components/index.html.md.erb
```diff
@@ -23,44 +27,18 @@ Read the [JavaScript API Reference](../javascript-api-reference/) for the list o
 
 The examples below follow our recommended [Import JavaScript using a bundler](../import-javascript/#import-javascript-using-a-bundler) approach and demonstrate how to:
 
-- [configure individual component instances](#configure-individual-component-instances)
+- [configure individual component instances](#configure-instances-of-a-specific-component)
 - [configure all components using the initAll function](#configure-all-components-using-the-initall-function)
```
36degrees, line 31:
```suggestion
- [configure all components using the `initAll` function](#configure-all-components-using-the-initall-function)
```

source/configure-components/index.html.md.erb
```diff
@@ -23,44 +27,18 @@ Read the [JavaScript API Reference](../javascript-api-reference/) for the list o
...
-
-```javascript
-import { CharacterCount } from 'govuk-frontend'
-
-const $element = document.querySelector('[data-module="govuk-character-count"]')
-new CharacterCount($element)
-```
-
-When initialised individually, errors are thrown rather than logged. You must check your application works without errors or some components will not work correctly.
+### Configure instances of specific components using the `createAll` function
```
romaricpascal, line 34: @36degrees Updated the headings so they both match. @claireashworth does that make the heading too long?

source/import-javascript/index.html.md.erb
```diff
@@ -65,23 +65,30 @@ If your service cannot use ES modules, read about [alternative module formats](#
 
 ### Import individual components
```
romaricpascal, line 66: I've updated the 'Configure components' to talk about 'specific components' rather than 'individual components' as we're now instanciating all instances of a given component rather than a single component (like we had in some of our previous examples). Should we align here as well?

source/configure-components/index.html.md.erb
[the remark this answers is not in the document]
romaricpascal: Done, we may want to check that [this doesn't make too long a heading](https://github.com/alphagov/govuk-frontend-docs/pull/443#discussion_r1596755476).

source/configure-components/index.html.md.erb
```diff
@@ -13,6 +13,10 @@ You can configure a component by:
 - [passing JavaScript configuration](#passing-javascript-configuration)
 - [adding HTML data attributes](#adding-html-data-attributes)
 
+Configuration passed via data attributes in the HTML or via Nunjucks macro options will take precedence over any JavaScript configuration.
```
claireashworth, line 16:
```suggestion
Configuration passed through data attributes in the HTML or Nunjucks macro options will take precedence over any JavaScript configuration.
```

romaricpascal: Updated 😊

claireashworth (approved):

PR #445, opened 2024-05-15
Expand documentation on importing specific bits of the Sass

36degrees:
Provide a more exhaustive example of how to import specific bits of our Sass, allowing users to understand what's possible without having to go digging through the code.

Specifically, document the full list of components and overrides as they comprise the greatest percentage of the generated CSS (70% and 15% respectively) and are least likely to be fully used.

I've also opted to include some approximate percentages to give users an indication of the weight of each layer. These percentages will be less accurate over time as the composition of our Sass changes, but I think they're likely to remain accurate enough to be useful.

👉🏻 [Have a look](https://deploy-preview-445--govuk-frontend-docs-preview.netlify.app/import-css/#import-specific-parts-using-sass) 👀

### DIFF
--- a/lib/package_contents.rb
+++ b/lib/package_contents.rb
@@ -1,13 +1,22 @@
 module PackageContents
   def components
-    File.readlines("node_modules/govuk-frontend/dist/govuk/components/_index.scss")
-      .map { |line| /@import "(?<component>[a-z-]*)\/index";/.match(line)&.[](:component) }
-      .compact
+    extract_from "components/_index.scss" do |line|
+      # Match anything found between `@import "` and `/index";`
+      line[/(?<=@import ")[a-z-]*(?=\/index";)/]
+    end
   end
 
   def overrides
-    File.readlines("node_modules/govuk-frontend/dist/govuk/overrides/_index.scss")
-      .map { |line| /@import "(?<override>[a-z-]*)";/.match(line)&.[](:override) }
+    extract_from "overrides/_index.scss" do |line|
+      # Match anything found between `@import "` and `";`
+      line[/(?<=@import ")[a-z-]*(?=";)/]
+    end
+  end
+
+  def extract_from(file, &block)
+    File.readlines("node_modules/govuk-frontend/dist/govuk/#{file}")
+      .map(&block)
       .compact
+      .tap { |list| raise "No matches found in #{file}" if list.empty? }
   end
 end
--- a/lib/package_contents.rb
+++ b/lib/package_contents.rb
@@ -1,0 +1,13 @@
+module PackageContents
+  def components
+    File.readlines("node_modules/govuk-frontend/dist/govuk/components/_index.scss")
+      .map { |line| /@import "(?<component>[a-z-]*)\/index";/.match(line)&.[](:component) }
+      .compact
+  end
+
+  def overrides
+    File.readlines("node_modules/govuk-frontend/dist/govuk/overrides/_index.scss")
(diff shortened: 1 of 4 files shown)


source/import-css/index.html.md.erb
```diff
@@ -40,20 +40,68 @@ To import all the Sass rules from GOV.UK Frontend, add the following to your Sas
 If you want to improve how quickly your service's pages load in browsers, you can import only the Sass rules you need.
 
 1. Import `node_modules/govuk-frontend/dist/govuk/base` in your Sass file.
-2. Import the parts of the CSS you need.
+2. Import only the Sass that you need, based on what components and other classes are being used in your service.
```
claireashworth, line 43:
```suggestion
2. Import only the Sass you need based on the components and other classes your service is using.
```

source/import-css/index.html.md.erb
```diff
@@ -40,20 +40,68 @@ To import all the Sass rules from GOV.UK Frontend, add the following to your Sas
...
 
-For example, add the following to your Sass file to import the CSS you need for a basic GOV.UK page.
+The biggest optimisations come from excluding any components or overrides that you aren't using.
```
claireashworth, line 45:
```suggestion
The biggest optimisations come from excluding any components or overrides you're not using.
```

source/import-css/index.html.md.erb
```diff
@@ -40,20 +40,68 @@ To import all the Sass rules from GOV.UK Frontend, add the following to your Sas
...
+
+You must import the layers in the order listed in the example below.
 
 ```scss
+// 'Base' includes everything from settings, tools and helpers. Nothing
+// in the base outputs any actual CSS.
```
claireashworth, line 51:
```suggestion
// in the base outputs any CSS.
```

source/import-css/index.html.md.erb
```diff
@@ -40,20 +40,68 @@ To import all the Sass rules from GOV.UK Frontend, add the following to your Sas
...
+// Alternatively, you can import all components:
+@import "node_modules/govuk-frontend/dist/govuk/components/index";
+*/
+
+// Utilities, for example govuk-clearfix or govuk-visually-hidden.
+// Approximately 1% of the CSS output if you include everything.
+
 @import "node_modules/govuk-frontend/dist/govuk/utilities/index";
+
+// Overrides, used to tweak things like the amount of spacing on an
```
claireashworth, line 83:
```suggestion
// Overrides, used to amend things like the amount of spacing on an
```

36degrees: Happy with 'adjust' instead? 'amend' doesn't feel natural in this context.

claireashworth (approved):

lib/package_contents.rb
```diff
@@ -0,0 +1,13 @@
+module PackageContents
+  def components
+    File.readlines("node_modules/govuk-frontend/dist/govuk/components/_index.scss")
+      .map { |line| /@import "(?<component>[a-z-]*)\/index";/.match(line)&.[](:component) }
```
36degrees, line 4: Thinking we should probably throw if we end up with an empty list? If we changed the structure or format of the index file (or moved to the Sass module system!) we'd want this to error rather than end up with an example with no components or overrides? 🤔

romaricpascal (approved):
Looking pretty neat! There's a chance a component from a commented out `@import` sneaks in, but after discussion on Slack, we think it's fairly unlikely so we can leave that for a later point (and either look at AST from the files or the filesystem).

PR #490, opened 2024-11-27
Tell users to use `index` instead of `all` when referring to sass imports

owenatgov:
What it says on the tin.

A companion PR to https://github.com/alphagov/govuk-frontend/pull/5518. Once we merge both of these we can close https://github.com/alphagov/govuk-frontend/issues/4988

## Notes
There are numerous references to `all` in the v4 portion of our docs. We shouldn't update these since v4 doesn't replace `all` with `index`.


source/import-css/index.html.md.erb
```diff
@@ -32,9 +32,11 @@ Then link the CSS file inside the `<head>` tag of your HTML page or page templat
 To import all the Sass rules from GOV.UK Frontend, add the following to your Sass file:
 
 ```scss
-@import "node_modules/govuk-frontend/dist/govuk/all";
+@import "node_modules/govuk-frontend/dist/govuk/index";
 ```
 
+You do not need `/index` at the end of your import if you’re using Dart Sass, LibSass 3.6.0 or higher, or Ruby Sass 3.6.0 or higher.
```
romaricpascal, line 38: 🙌🏻  Makes sense to be consistent with the rest of the site about this. I'll create an [issue to remove statements](https://github.com/alphagov/govuk-frontend-docs/issues/491) about LibSass 3.6.0 and Ruby Sass 3.6.0 cross site.

source/installing-with-npm/index.html.md.erb
```diff
@@ -66,7 +66,7 @@ The accordion will use a generic font until you get the font and images working,
 There are also different ways you can [import GOV.UK Frontend's CSS](../import-css/), including into your project's main Sass file:
 
     ```scss
-    @import "node_modules/govuk-frontend/dist/govuk/all";
+    @import "node_modules/govuk-frontend/dist/govuk/index";
     ```
```
romaricpascal, line 70: **suggestion** For consistency with the other examples about `index` files 😊 

```suggestion
    @import "node_modules/govuk-frontend/dist/govuk/index";
    ```

You do not need `/index` at the end of your import if you’re using Dart Sass, LibSass 3.6.0 or higher, or Ruby Sass 3.6.0 or higher.
```

romaricpascal (approved):
Thanks for raising the PR on the Frontend Docs 🙌🏻  Nothing major, just a little suggestion 😊 Doesn't change the tech side of things, so I'll approve.

source/import-css/index.html.md.erb
owenatgov: I wasn't sure about this one. We want to encourage folks to not use deprecated packages but also give folks the correct info if they are and aren't in a position to move off them. I feel like these docs lean more towards the latter.

source/installing-with-npm/index.html.md.erb
owenatgov: I thought about doing this but wondered if the link to the sass importing docs above carried that context. It can't hurt to put it here though so I'll commit this.

PR #510, opened 2025-03-25
Remove pre-release installation instructions

domoscargin:
We don't do pre-releases very often, so the version behind the "next" tag is usually pretty out of date.

We can encourage folks to try the "next" tag via comms when we need wider testing. Most of the active port maintainers are already pretty on top of this anyway.


owenatgov (approved):
Get it outta here!

querkmachine (approved):

PR #522, opened 2025-05-21
Draft changes to tech docs for v4 brand update

CharlotteDowns:
## What 
Update to Changes to GOV.UK Frontend v5.0.0 to reflect the work we're doing to support adopting the GOV.UK rebrand.

## Why
Users of v4 and v5 are able to understand the context of us supporting an update to the rebrand. We also include that we are not planning to support any future updates to v4.


owenatgov (commented):
Looks good from a dev perspective. I'll let our content pals have final say and approval once the release notes are finished

PR #524, opened 2025-05-30
Update outdated link to plural rules chart

36degrees:
The link https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html is outdated – it redirects to the homepage for the staging site with a banner that says 'We've Moved! Please update your links.'

I believe I've found [the equivalent page on the new domain](https://www.unicode.org/cldr/charts/47/supplemental/language_plural_rules.html). Preserving the `latest` in the URL redirects to the chart as it exists in the latest CLDR version.


romaricpascal (approved):
🦅 Good spot! Looks like the right chart to point to 🙌🏻

PR #553, opened 2025-09-04
Fix typo in error handler context and other suggestions

colinrotherham:
This PR aimed to fix a ~`Component`~ `component` typo in the error handler example:

```patch
- function notifyErrorMonitoringService(error, { Component, element, config }) {
+ function notifyErrorMonitoringService(error, { component, element, config }) {
```

But I also noticed an incorrect link to copying assets and some other minor improvements

### DIFF
--- a/source/building-your-own-javascript-components/index.html.md.erb
+++ b/source/building-your-own-javascript-components/index.html.md.erb
@@ -218,7 +218,7 @@ The context object will contain the following properties:
 import { createAll } from 'govuk-frontend'
 import { ProjectComponent } from './project-component.js'
 
-function notifyErrorMonitoringService(error, { Component, element, config }) {
+function notifyErrorMonitoringService(error, { component, element, config }) {
   // Send relevant details to an error monitoring service
 }
 
@@ -238,7 +238,7 @@ import { ProjectComponent } from './project-component.js'
 
 const $element = document.querySelector('.app-dialog')
 
-function notifyErrorMonitoringService(error, { Component, element, config }) {
+function notifyErrorMonitoringService(error, { component, element, config }) {
   // Do something with initialisation error, like sending relevant details to an error monitoring service
 }
 
--- a/source/building-your-own-javascript-components/index.html.md.erb
+++ b/source/building-your-own-javascript-components/index.html.md.erb
@@ -387,7 +387,7 @@ class MyComponent extends ConfigurableComponent {
       // As `anOption` is left out of this object
       // it won’t be able to be overridden by `data-an-option`
       aNumberOption: {type: 'number'},
-      aBooleanOption: {type: 'boolean'}
+      aBooleanOption: {type: 'boolean'},
       // For nested objects, only the type of the root needs to be defined
       anOptionWithNesting: {type: 'object'}
     }
--- a/source/building-your-own-javascript-components/index.html.md.erb
+++ b/source/building-your-own-javascript-components/index.html.md.erb
@@ -386,10 +386,10 @@ class MyComponent extends ConfigurableComponent {
     properties: {
       // As `anOption` is left out of this object
       // it won’t be able to be overridden by `data-an-option`
-      aNumberOption: {type: 'number'},
-      aBooleanOption: {type: 'boolean'},
+      aNumberOption: { type: 'number' },
+      aBooleanOption: { type: 'boolean' },
       // For nested objects, only the type of the root needs to be defined
-      anOptionWithNesting: {type: 'object'}
+      anOptionWithNesting: { type: 'object' }
     }
   }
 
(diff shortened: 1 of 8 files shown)


source/installing-with-npm/index.html.md.erb
```diff
@@ -112,7 +112,7 @@ In your live application, we recommend [using an automated task or your build pi
 
 In your live application, we recommend:
 
-- [using an automated task or your build pipeline](../import-javascript/) instead of copying the files manually
+- [using an automated task or your build pipeline](../import-font-and-images-assets/) instead of copying the files manually
```
colinrotherham, line 115: The build pipeline link text is used in various places, but this one has the wrong link

source/v4/importing-css-assets-and-javascript/index.html.md.erb
```diff
@@ -214,7 +214,7 @@ GOVUKFrontend.initAll()
 
 ### Select and initialise part of a page
 
-If you update a page with new markup, for example a modal dialogue box, you can run `initAll` with a `scope` parameter to initialise the components on part of a page.
+If you update a page with new markup, for example a modal dialogue box, you can run `initAll` with a `{ scope }` option to initialise the components on part of a page.
```
colinrotherham, line 217: I've clarified how the words "option", "parameter" and "property" are used

Looks like `initAll({ scope })` using options has always been the default?

Since running `initAll($scope)` as a parameter was only added in: https://github.com/alphagov/govuk-frontend/pull/6164

source/javascript-api-reference/index.html.md.erb
```diff
@@ -266,7 +266,7 @@ Default:
 
 Type: object
 
-Message made available to assistive technologies, if none is already present in the HTML, to describe that the component accepts only a limited amount of content. The component will replace the `%{count}` placeholder with the value of the `maxlength` or `maxwords` parameter.
+Message made available to assistive technologies, if none is already present in the HTML, to describe that the component accepts only a limited amount of content. The component will replace the `%{count}` placeholder with the value of the `maxlength` or `maxwords` option.
```
colinrotherham, line 269: The docs always say `html` or `text` "option" but here it was "parameter"

source/building-your-own-javascript-components/index.html.md.erb
```diff
@@ -238,14 +245,15 @@ import { ProjectComponent } from './project-component.js'
 
 const $element = document.querySelector('.app-dialog')
 
-function notifyErrorMonitoringService(error, { Component, element, config }) {
+function notifyErrorMonitoringService(error, { component, element, config }) {
```
colinrotherham, line 248: The error context `component` property is lowercase

source/import-javascript/index.html.md.erb
```diff
@@ -73,28 +73,43 @@ To improve how quickly your service’s pages load in browsers, import only the
...
 import { SkipLink, CharacterCount, createAll } from 'govuk-frontend'
 
-createAll(SkipLink, function(error) {
+createAll(SkipLink, (error, context) => {
   // Handle the error here, for example send it to an error monitoring service
 })
+
 // You can provide a config for components that use them
-createAll(CharacterCount, { maxLength: 500 }, function(error) {
+createAll(CharacterCount, { maxLength: 500 }, (error, context) => {
```
colinrotherham, line 95: The 2nd context parameter was missing here

source/building-your-own-javascript-components/index.html.md.erb
```diff
@@ -386,10 +394,10 @@ class MyComponent extends ConfigurableComponent {
     properties: {
       // As `anOption` is left out of this object
       // it won’t be able to be overridden by `data-an-option`
-      aNumberOption: {type: 'number'},
-      aBooleanOption: {type: 'boolean'}
+      aNumberOption: { type: 'number' },
+      aBooleanOption: { type: 'boolean' },
```
colinrotherham, line 398: There was a missing comma here 😬

source/installing-with-npm/index.html.md.erb
36degrees: I think the original link was correct – this section is about 'Getting the JavaScript working'.

In step 2 we tell users to copy the minified JavaScript into their application. This is then saying that actually we'd rather they used an automated task to include the JS from the package, rather than copying it like this.

36degrees (changes requested):
Thanks for taking the time to put this together, @colinrotherham 🙏🏻 

I think the changes broadly look fine, except I don't think the link change in the second commit is correct.

I would like us to look at moving some of the documentation about handling errors outside of the 'Creating your own components' guidance because it doesn't feel like the right place for it, but I don't think we should tackle that as part of this PR.

This'll also need a second review before we can merge it – ideally from @seaemsi if possible.

36degrees:
Yep sure, I can drop 98a8b3b2f767984ba138d7468c9799d02b5edd11 on your behalf when I get a sec.

Appreciate the small commits and review comments, made it nice and easy to review.

Issue #586, opened 2026-01-19
Create a 6.0.0 'version' of the frontend docs


owenatgov:
## What

Clone the current frontend docs into a 'v5' subfolder, create a new top level right hand navigation option of '6.x (latest)' and remove, edit and/or update docs pertaining to v5 to instead reference v6. We did this when we released v5 for [v4](https://frontend.design-system.service.gov.uk/v4/).

## Why

So that it's easier for users on an upgrade path from old versions of Frontend or ancestors to Frontend to move up through the major versions.

## Done when

- [x] New sub folder created
- [ ] v5 content adjusted to identify it as no longer latest
- [ ] References to v5 removed from v6/current content

kellylee-gds:
@owenatgov comparing [the Padlet card](https://govukdesignsystem.padlet.org/govukdesignsystem/v6-what-s-left-to-do-e4ar4lufvkiprm3f/wish/LNV1Q70R2v6OZmq3) to the issue, I'm wondering if there are a few things we should call out explicitly which are covered by a v6 version of the FE docs?

They are:
- update guidance for updating from legacy versions
- update benefits of updating

Do you think it makes sense to cover these in this card, or are they distinct enough to be their own cards?

kellylee-gds:
@owenatgov happy with that - let's take an mvp approach for now.

PR #587, opened 2026-01-19
Add a v5 version of the frontend docs

owenatgov:
Resolves https://github.com/alphagov/govuk-frontend-docs/issues/586

Specific changes are:

- adds the v5 section the same way we did with v4 ie: new nav item, new discrete v5 package for generating a distinct sass doc
- adjusts the index in the new v5 section to try to point people towards upgrading
- redirects the 'changes in v5' page

To simplify review, the only change I made to the copied content into the v5 section is the index at the root of the folder. Everything else is the same as what's currently live.

## Thoughts

This does the job for now but the approach is ultimately a plaster. It's going to run out of steam in terms of room in the nav bar and how many different sassdocs we're generating eventually. I suggest we put some time in the roadmap to reconsider the whole journey of these docs, including the design and content strategy of these old versions.

The content changes I made here are very very rough and I would appreciate thoughts on the substance of them first, content quaility second.

### DIFF
--- a/source/brand-refresh-changes/index.html.md.erb
+++ b/source/brand-refresh-changes/index.html.md.erb
@@ -7,36 +7,16 @@ weight: 4
 
 In June 2025, GOV.UK started to refresh the brand across its products and services.
 
-The changes affect these components:
+## If you’re using GOV.UK Frontend v6 or later
 
-- [GOV.UK header](https://design-system.service.gov.uk/components/header/)
-- [GOV.UK footer](https://design-system.service.gov.uk/components/footer/)
-- [Cookie banner](https://design-system.service.gov.uk/components/cookie-banner/)
-- [Service navigation](https://design-system.service.gov.uk/components/service-navigation/)
+You do not need to take any action. Your service is using the refreshed GOV.UK branding.
 
-## If you’re using GOV.UK Frontend v5
+## If you’re using GOV.UK Frontend v5 or earlier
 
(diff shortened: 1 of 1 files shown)


owenatgov:
@romaricpascal done. This PR doesn't include any sort of 'changes in v6' page or a brand refresh page at root.

Issue created for a changes in v6 page https://github.com/alphagov/govuk-frontend-docs/issues/592

I'm gonna leave the brand page until we've chatted about this PR later today. Depending on outcomes, we might just leave it as it is in this PR.

source/brand-refresh-changes/index.html.md.erb
```diff
@@ -7,36 +7,12 @@ weight: 4
...
 In June 2025, GOV.UK started to refresh the brand across its products and services.
 
-The changes affect these components:
+## If you’re using GOV.UK Frontend v6 or later
 
-- [GOV.UK header](https://design-system.service.gov.uk/components/header/)
-- [GOV.UK footer](https://design-system.service.gov.uk/components/footer/)
-- [Cookie banner](https://design-system.service.gov.uk/components/cookie-banner/)
-- [Service navigation](https://design-system.service.gov.uk/components/service-navigation/)
+You don't need to take any action. The refreshed brand is applied by default.
```
calvin-lau-sig7:
```suggestion
You do not need to take any action. Your service is using the refreshed GOV.UK branding.
```

PR #601, opened 2026-02-20
Remove trailing `/index` of `@import` statements in v6.0.0 docs

### LINKED ISSUE #600: Remove `/index` from code examples in Frontend docs
## What

Remove the trailing `index` from code examples in Frontend docs

## Why

Since we now only support Dart Sass, we can make the code we recommend less verbose. This is because Dart Sass will automatically look for `path/to/somewhere/index.scss` or `path/to/somewhere/_index.scss` when resolving  `path/to/somewhere` with `@import` , `@use` or `@forward`. 

## Who needs to work on this

Tech writer, developers

## Who needs to review this

Developers

## Done when

- [x] We've removed all trailing `index` files from the code examples in GOV.UK Frontend

romaricpascal:
v6.0.0 only supports Dart Sass 1.79+, which will automatically loads the `index.scss` or `_index.scss` files if the path resolves to a folder.

Only the content of the v6.0.0 docs has been updated, as v5.0.0 and v4.0.0 still support Ruby Sass and LibSass which would need the full path to the imported file.

Fixes #600

### DIFF
--- a/source/import-css/index.html.md.erb
+++ b/source/import-css/index.html.md.erb
@@ -55,36 +55,36 @@ You must import the layers in the order listed in the example below.
 // Basic content styles for typography, links etc. Approximately 10% of
 // the CSS output if you include everything.
 
-@import "node_modules/govuk-frontend/dist/govuk/core/index";
+@import "node_modules/govuk-frontend/dist/govuk/core";
 
 // Objects include things like the page template, grid and form groups.
 // Approximately 5% of the CSS output if you include everything.
 
-@import "node_modules/govuk-frontend/dist/govuk/objects/index";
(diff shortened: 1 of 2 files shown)


romaricpascal:
> Just a note that we should also be doing something about our [`components` def in PackageContents](https://github.com/alphagov/govuk-frontend-docs/blob/78b709bda1fbda642decda14ca29fd92e470d7b1/lib/package_contents.rb#L2-L7) at some point.

Yeah, noted that on Slack on Friday. I think it's part of an issue to remove `/index` inside GOV.UK Frontend itself. I'll draft that shortly 😊

PR #615, opened 2026-04-23
6.2.0 content changes

### LINKED ISSUE #620: Add documentation updates for sass modules
## What
Add changes from [this internal doc](https://docs.google.com/document/d/1oBihWFyuieqa6lE-ti6x7COYxr5lHYU5KOkxrQqawj4/edit?tab=t.c7gzwmw153mg) to the docs website.

## Why
So users can read about how to `@use` our code.

Technical writer, developers

## Done when

- [ ] The docs are live on the [#615 preview deployment](https://deploy-preview-615--govuk-frontend-docs-preview.netlify.app/)

romaricpascal:
PR for gathering the content changes for the 6.2.0 release that'll bring in our use and support for Sass module.

The branch currently runs against a preview release of [the `feature/sass-modules` branch of GOV.UK Frontend](https://github.com/alphagov/govuk-frontend/pull/6862).

[Preview of the Sass API reference](https://deploy-preview-615--govuk-frontend-docs-preview.netlify.app/sass-api-reference/) - [Original Sass API reference](https://frontend.design-system.service.gov.uk/sass-api-reference/) (note that the disparition of `Objects` is expected as [the `govuk-width-container` mixin has been moved to helpers](https://github.com/alphagov/govuk-frontend/pull/6946)).

- https://github.com/alphagov/govuk-frontend-docs/pull/616


romaricpascal:
The Sass API compiled OK, the [`govuk-width-container `mixin is appropriately listed in the helpers section now](https://github.com/alphagov/govuk-frontend/pull/6946).

Only hiccup seems to be the duplication of the settings for the buttons, I think because they're documented in both the `components/button/_settings.scss` and `components/button/_settings.import.scss`. I propose we only document them in the module, and not in the import-only file.

On the bright side, all the new internal mixins to support both `@import` and `@use` are not appearing in the Sass Doc (but we should still explicitly mark them private for clarity) 🎉 

[Diff of the HTML from the two pages](https://www.diffchecker.com/oTePn0NT/)

NickColley (approved):

PR #616, opened 2026-04-24
Exclude import-only files from Sassdoc

### LINKED ISSUE #617: Create a preview release for use with Frontend Docs and audit sassdoc output
## Done when

- [ ] Thing to do

romaricpascal:
v6.2.0 adds import only files in GOV.UK Frontend, some of which have duplicate the Sass documentation from their corresponding Sass module. To avoid duplicating entries in the Sass API, we use a `sassdoc.config.json` file to exclude the `.import.scss` file.

Unfortunately, there's [no flag for Sassdoc's CLI to exclude files](http://sassdoc.com/getting-started/#options) and working out a glob pattern would make things a little cryptic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,9 @@
   "scripts": {
     "start": "bundle exec middleman server",
     "postinstall": "npm run build:sassdoc && npm run build:sassdocv5 && npm run build:sassdocv4",
-    "build:sassdoc": "sassdoc --no-update-notifier --parse node_modules/govuk-frontend/dist/govuk/ > data/sassdoc.json",
-    "build:sassdocv5": "sassdoc --no-update-notifier --parse node_modules/govuk-frontend-v5/dist/govuk/ > data/sassdoc-v5.json",
-    "build:sassdocv4": "sassdoc --no-update-notifier --parse node_modules/govuk-frontend-v4/govuk/ > data/sassdoc-v4.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "bundle exec middleman server",
     "postinstall": "npm run build:sassdoc && npm run build:sassdocv5 && npm run build:sassdocv4",
-    "build:sassdoc": "sassdoc --no-update-notifier --parse node_modules/govuk-frontend/dist/govuk/ > data/sassdoc.json",
+    "build:sassdoc": "sassdoc --no-update-notifier --parse node_modules/govuk-frontend/dist/govuk/ -c sassdoc.config.json > data/sassdoc.json",
```
romaricpascal: **note** Only the v6 sassdoc needs to exclude the import only files, so I've only added the configuration there. Happy to extend it to the rest of the `build:sassdoc` commands if we prefer consistency.

36degrees (commented):
Wondering if it'd be worth:

- moving `no-update-notifier` into the config file, so we pass all the non input/output config the same way
- updating all the `build:sassdoc*` scripts to use the config – this'd mean v4 and v5 would also exclude import-only files, but there shouldn't be any, so that's OK?

36degrees (approved):

Issue #618, opened 2026-04-24
Comments in code blocks are almost invisible


36degrees:
## Description of the issue

Comments in code blocks are now #eee on #f9f9f9, with a color contrast of 1.1:1, and are pretty much invisible

<img width="1624" height="1061" alt="Image" src="https://github.com/user-attachments/assets/d6ed4d81-f9c2-4ea3-b145-e16bb8255110" />

I'm pretty sure this is a recent regression, possibly introduced as part of an update to the tech docs gem.

## Steps to reproduce the issue

1. Go to https://deploy-preview-615--govuk-frontend-docs-preview.netlify.app/import-css/#import-specific-parts-using-sass
2. Look at the code block
3. Look even closer – I promise there are comments there

## Actual vs expected behaviour

You should be able to read the comments.

NickColley:
The immediate issue is resolved by #621 but will leave this open until the fix is upstreamed into the tech-docs-gem

PR #621, opened 2026-04-28
Fix contrast issues with comments in code examples

NickColley:
The colours were updated to better [match version 6 but for comments they were mistakenly set to the wrong colour](https://github.com/alphagov/tech-docs-gem/pull/442/changes/f09a4b8673fef2150e7aca18c4cc120774329656#diff-81435f36c443c406a3436b0531a4991cd20655917ede750cec10d5b2c6b3bdcf).

This results in comments that are very hard to read.

I have put in a quick patch here to resolve the issue immediately and will upstream this patch into the tech-docs-gem.

### Before
Screenshot of code example with comments that have very low contrast 

### After
Screenshot of code example with comments that have very low contrast

### DIFF
--- a/source/stylesheets/screen.css.scss
+++ b/source/stylesheets/screen.css.scss
@@ -4,3 +4,12 @@
 .technical-documentation .table-container td > code {
   white-space: nowrap;
 }
+
+// Temporarily override syntax highlighting to fix poor contrast for comments.
+.highlight .c,
+.highlight .cm,
+.highlight .cp,
+.highlight .c1,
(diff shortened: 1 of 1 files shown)


domoscargin (approved):
Looks good - can we add an issue to remove this fix when we bump tech docs?

NickColley:
Will do thank you!

NickColley:
Ah already raised:

https://github.com/alphagov/govuk-frontend-docs/issues/618

PR #625, opened 2026-05-26
Warn about using `@import` with `pkg:` URLs

romaricpascal:
Sass won't load the import-only files from GOV.UK Frontend, so `pkg:` URLs are best reserved to `@use`.

### DIFF
--- a/source/include-css/index.html.md.erb
+++ b/source/include-css/index.html.md.erb
@@ -56,7 +56,7 @@ If you’re using a bundler such as Webpack or Vite, it may already locate GOV.U
 
 ### Use `pkg:` URLs to shorten Sass URLs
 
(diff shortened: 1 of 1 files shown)


calvin-lau-sig7 (dismissed):
Looks fine to me, it reads well and flows into the "If you cannot use..." paragraph.

NickColley (dismissed):
Nice 👍🏻

romaricpascal:
Updated with the same warning as in the [CHANGELOG](https://github.com/alphagov/govuk-frontend/pull/7107/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR61-R68). Given we have access to GOV.UK Frontend component's render a proper Inset text component instead of Markdown quotes.

Unfortunately, requires the content to be in HTML rather than Markdown as the markdown renderer didn't transform the content between `<div class="govuk-inset-text>` and `</div>`, even when adding double line breaks.

NickColley (approved):

owenatgov (approved):

PR #660, opened 2026-09-11
Check and fix external links

36degrees:
Add a new `check-links:external` script that runs linkinator against external links.

Unlike `check-links` I haven't set this up to run on GitHub Actions – it's too flakey, with GitHub in particular often returning 429s. It might be that with the right configuration (e.g. `retry`) we could set it up to run automatically, but I'm leaving that as a future enhancement.

See individual commits for further details.


source/v4/migrating-from-legacy-products/index.html.md.erb
```diff
@@ -72,7 +72,7 @@ You should also follow our [guidance if you extend or modify GOV.UK Frontend com
 We recommend migrating in the following order:
 
 - [components](https://design-system.service.gov.uk/components/)
-- [typography](https://design-system.service.gov.uk/styles/typography/)
+- [typography(https://design-system.service.gov.uk/styles/)
```
NickColley: I think a bracket missing?

NickColley (changes requested):

source/v4/migrating-from-legacy-products/index.html.md.erb
36degrees: I don't really have a better idea as there's no single landing place for typography. My hope is that given it's clearly marked out as a section in the nav on https://design-system.service.gov.uk/styles/ that it'd suffice, at least for now?

NickColley (approved):

PR #662, opened 2026-09-11
Fix GitHub source urls in Sass API reference

### LINKED ISSUE #657: Sass API reference for previous major versions links to incorrect source file
## Description of the issue

The 'View source code' links in the Sass API reference for [v4](https://frontend.design-system.service.gov.uk/v4/sass-api-reference/) and [v5](https://frontend.design-system.service.gov.uk/v5/sass-api-reference/) reference the current v6 version rather than the latest v4.x or v5.x release in their URLs.

In some cases these links 404 because they have been moved between v4 and v5. In most cases the fragments that link to the relevant lines in the file either don't exist, or highlight the wrong part of the file.

## Steps to reproduce the issue

1. View the Sass API Reference for [v4](https://frontend.design-system.service.gov.uk/v4/sass-api-reference/) and [v5](https://frontend.design-system.service.gov.uk/v5/sass-api-reference/)
2. Examine the 'View source' links and see that they link to GitHub using the latest (e.g. v6) tag

## Actual vs expected behaviour

For example, the 'View source' link for `$govuk-assets-path` in the v5 documentation links to:

```
https://github.com/alphagov/govuk-frontend/tree/v6.5.0/packages/govuk-frontend/src/govuk/settings/_assets.scss#L14-L14
```

When it should link to:
```
https://github.com/alphagov/govuk-frontend/tree/v5.14.0/packages/govuk-frontend/src/govuk/settings/_assets.scss#L14-L14
```

## Environment (where applicable)

N/A

36degrees:
An attempt to fix #657.

Not entirely sure whether this is the right approach, so raising as a draft to get feedback.


NickColley:
Not a Ruby developer really so cant comment on the helper approach but it is easy to understand even with that in mind and solves the problem so seems good to me 👍🏻

